PR #24, opened 2015-01-15
Show correct subject in exception message

dantleech:
The Assert:path currenly shows:

````
The path "/" is not absolute.
````

when it should say

````
The path "foo/bar" is not absolute.
````


dantleech:
I couldn't find any tests for this class .. I can create one if you want.

webmozart:
The class is currently only tested indirectly through other classes. If you create some, that's perfect! :)

dantleech:
Added tests for the two cases that I fixed.

webmozart:
Thanks! :)

PR #25, opened 2015-02-28
Show mount points when listing composite repositories

dantleech:
This PR is an incomplete POC and aims to enable the listing of mount points in composite repositories.

For example if I do:

````php
$compRepo->mount('/foobar', $fooRepo);
$compRepo->mount('/foobar/barfoo', $barfooRepo);
$compRepo->mount('/dab', $dabRepo);
````

Then I would expect to see `foobar` and `dab` when doing:

````php
$compRepo->listChildren('/'); // show both foobar and dab as children
````

In addition, retrieving the root resource (`/`) when no mountpoint has been set (at `/`) will return a `CompositeResource` resource.

Does it make sense?

### DIFF
--- a/src/CompositeRepository.php
+++ b/src/CompositeRepository.php
@@ -18,6 +18,7 @@ use Puli\Repository\Api\ResourceRepository;
 use Puli\Repository\Assert\Assert;
 use Puli\Repository\Resource\Collection\ArrayResourceCollection;
 use Webmozart\PathUtil\Path;
+use Puli\Repository\Resource\CompositeResource;
 
 /**
  * A repository combining multiple other repository instances.
@@ -132,6 +133,12 @@ class CompositeRepository implements ResourceRepository
      */
     public function get($path)
     {
+        if ('/' === $path) {
+            $composite = new CompositeResource('/');
+            $composite->attachTo($this);
+            return $composite;
+        }
+
         list ($mountPoint, $subPath) = $this->splitPath($path);
 
         if (null === $mountPoint) {
@@ -182,6 +189,10 @@ class CompositeRepository implements ResourceRepository
      */
     public function hasChildren($path)
     {
+        if ('/' === $path) {
+            return count($this->repos) ? true : false;
+        }
+
         list ($mountPoint, $subPath) = $this->splitPath($path);
 
         if (null === $mountPoint) {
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: src/CompositeRepository.php [hunk 1]
```diff
@@ -182,6 +189,10 @@ public function contains($query, $language = 'glob')
      */
     public function hasChildren($path)
     {
+        if ('/' === $path) {
+            return count($this->repos) ? true : false;
```
dantleech: This is wrong.

src/CompositeRepository.php
```diff
@@ -199,8 +210,27 @@ public function hasChildren($path)
...
     public function listChildren($path)
     {
+        $compositeCollection = new ArrayResourceCollection();
+
+        foreach ($this->repos as $mountPointPath => $repo) {
+            if (0 !== strpos($mountPointPath, $path)) {
+                continue;
+            }
+
+            if (substr_count($mountPointPath, '/') !== 1) {
```
dantleech: Should be the remaining path after subtracting the requested path.

src/CompositeRepository.php
```diff
@@ -199,8 +210,27 @@ public function hasChildren($path)
...
+            }
+
+            $rootResource = $repo->get('/');
+            $compositeCollection->add($rootResource->createReference($mountPointPath));
+        }
+
         list ($mountPoint, $subPath) = $this->splitPath($path);
 
+        if (null === $mountPoint && '/' === $path) {
+            return $compositeCollection;
```
dantleech: Return the composite collection if nothing is mounted on `/`

src/CompositeRepository.php
```diff
@@ -208,10 +238,14 @@ public function listChildren($path)
...
         }
 
-        $resources = $this->getRepository($mountPoint)->listChildren($subPath);
-        $this->replaceByReferences($resources, $mountPoint);
+        $collection = $this->getRepository($mountPoint)->listChildren($subPath);
+        $this->replaceByReferences($collection, $mountPoint);
 
-        return $resources;
+        foreach ($collection as $resource) {
+            $compositeCollection->add($resource);
```
dantleech: The mounted resources should probably override any existing resources, this doesn't do that.

src/CompositeRepository.php
```diff
@@ -235,7 +269,9 @@ private function splitPath($path)
                     return array($mountPoint, $path);
                 }
 
-                return array($mountPoint, substr($path, strlen($mountPoint)));
+                $subPath = substr($path, strlen($mountPoint)) ? : '/';
```
dantleech: This should probably be in a different PR. It fixes the issue when trying to access the root resource of composite respources.

webmozart:
Thank you for that PR! I will process it once the development of version 1.1 starts, as the CompositeRepository will not be part of 1.0.

dantleech:
Its actually a pretty terrible PR :) But would be happy to have another go if you agree with the behavior

tgalopin:
I think this can be closed as the PathMappingRepository now implement this.

PR #27, opened 2015-06-25
Implement OptimizedPathMappingRepository

tgalopin:
As explained in puli/issues#80, I implemented an optmized version of the PathMappingRepository. This pull-request is not ready for the moment, please do not merge it now. It's just a place for dicussion about the code.

I want to explain what I did and what I plan to do to discuss, especially with @webmozart :) .

1. Create the repository ✓
---------------------------------------------------------------------------------------

https://github.com/puli/repository/compare/1.0...tgalopin:path-mapping?diff=unified&name=path-mapping#diff-55bbc453f8c7d643ba2cd09cc280c00f

As discussed with @webmozart, optimized and not optimized versions of the PathMappingRepository are completely different in code, it makes sense to separate them. I created the optimized version only for the moment and I'll open another pull-request for the classical one.

2. Create a GenericFilesystemResource ✓
---------------------------------------------------------------------------------------

https://github.com/puli/repository/compare/1.0...tgalopin:path-mapping?diff=unified&name=path-mapping#diff-2ba3bf2e348a04a316173aba204de2d9

It's the best solution I found to store a virtual directory in the repository. Here is the problem:

- I want to store only FilesystemResource instances in the repository ;
- However, when someone add the path `/webmozart/puli/file1`, the paths `/webmozart` and  `/webmozart/puli` should be created if they does not exist. But they don't have to represent a real file on the system ;

So I created this class to represent a generic, virtual filesystem resource usable in the repository while not having any real file / directory behind it. Perhaps are there better ways or just a better name for this class, please tell me if you think of something :) .

3. Externalize test resources creation in order to be able to override it ✓
---------------------------------------------------------------------------------------

https://github.com/puli/repository/compare/1.0...tgalopin:path-mapping?diff=unified&name=path-mapping#diff-59ad75cfaee83a692af80d3634224473

As my repository only handle FilesystemResource instances, I needed to create specific testing objects. I had to externalize this creation, so I edited the abstract tests in order to do so.

4. Add specific tests for the optimized version ✓
---------------------------------------------------------------------------------------

https://github.com/puli/repository/blob/a25287982a5d2c64c0ed507532110af89bd614b8/tests/OptimizedPathMappingRepositoryTest.php

Test that the adding of a single directory does add all the children of this directory in the right place.

Do you think of other specific test I could have forgotten?


src/OptimizedPathMappingRepository.php
```diff
@@ -0,0 +1,365 @@
...
+        $paths = iterator_to_array($iterator);
+        $resources = array_values($this->store->getMultiple($paths));
+
+        return new FilesystemResourceCollection($resources);
+    }
+
+    /**
+     * Sort the store by keys
+     */
+    private function sortStore()
```
tgalopin: I didn't find a better way to sort the store. Does someone have a proposal?

src/OptimizedPathMappingRepository.php
```diff
@@ -0,0 +1,365 @@
...
+    {
+        $resources = $this->store->getMultiple($this->store->keys());
```
tgalopin: I didn't find a better way to sort the store. Does someone have a proposal?

src/OptimizedPathMappingRepository.php
```diff
@@ -0,0 +1,365 @@
...
+    }
+
+    /**
+     * Returns an iterator for the children of a resource.
+     *
+     * @param FilesystemResource $resource The resource.
+     *
+     * @return RegexFilterIterator|Resource[] The iterator.
+     */
+    private function getChildIterator(FilesystemResource $resource)
```
tgalopin: This function (and `getGlobterator()`) returns keys where equivalent in other repositories returns resources. It does not seem to be a problem as the method is private, but is it?

src/OptimizedPathMappingRepository.php
```diff
@@ -0,0 +1,365 @@
...
+        }
+    }
+
+    /**
+     * Transform an iterator of paths into a collection of resources
+     *
+     * @param \Iterator $iterator
+     * @return FilesystemResourceCollection
+     */
+    private function iteratorToCollection(\Iterator $iterator)
```
tgalopin: As `getGlobalIterator()` and `getChildIterator()` return paths, this method transform these path into a collection of resources.

src/OptimizedPathMappingRepository.php
```diff
@@ -0,0 +1,365 @@
...
+            throw UnsupportedLanguageException::forLanguage($language);
+        }
+
+        Assert::stringNotEmpty($query, 'The glob must be a non-empty string. Got: %s');
+        Assert::startsWith($query, '/', 'The glob %s is not absolute.');
+
+        $query = Path::canonicalize($query);
+        $resources = array();
+
+        if (false !== strpos($query, '*')) {
```
webmozart: You should use `Glob::isDynamic()` instead, as globs can contain other dynamic parts than `*` (e.g. `{foo,bar}`).

src/OptimizedPathMappingRepository.php
```diff
@@ -0,0 +1,365 @@
...
+        if ('glob' !== $language) {
+            throw UnsupportedLanguageException::forLanguage($language);
+        }
+
+        Assert::stringNotEmpty($query, 'The glob must be a non-empty string. Got: %s');
+        Assert::startsWith($query, '/', 'The glob %s is not absolute.');
+
+        $query = Path::canonicalize($query);
+
+        if (false !== strpos($query, '*')) {
```
webmozart: Same here

src/OptimizedPathMappingRepository.php
```diff
@@ -0,0 +1,365 @@
...
+
+        return $nbOfResources - count($this->store->keys());
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function clear()
+    {
+        $root = new GenericFilesystemResource('/');
```
webmozart: Why can't us use GenericResource?

src/OptimizedPathMappingRepository.php
```diff
@@ -0,0 +1,365 @@
...
+        $root->attachTo($this, '/');
+
+        // Subtract root
+        $removed = count($this->store->keys()) - 1;
```
webmozart: I'd change this to

```php
$removed = $this->store instanceof Countable
    ? count($this->store) - 1
    : count($this->store->keys()) - 1;
```

so that stores can provide optimized counting.

src/OptimizedPathMappingRepository.php
tgalopin: Here I could, but in the adding process (addResource()) I can't as the repository need instances of FilesystemRepository. So I thought as I used it in addResource(), why not use it here :) .

src/OptimizedPathMappingRepository.php
```diff
@@ -0,0 +1,365 @@
...
+
+    /**
+     * @param FilesystemResource $resource
+     */
+    private function removeResource(FilesystemResource $resource)
+    {
+        $path = $resource->getPath();
+
+        // Ignore non-existing resources
+        if (! $this->store->exists($path)) {
```
webmozart: Please omit spaces between `!` and variables.

src/OptimizedPathMappingRepository.php
tgalopin: I just copied it from InMemoryRepository, should I change it there too?

src/OptimizedPathMappingRepository.php
tgalopin: Okay, no problem :)

src/OptimizedPathMappingRepository.php
tgalopin: Good idea :)

src/OptimizedPathMappingRepository.php
webmozart: No that's fine :)

src/OptimizedPathMappingRepository.php
```diff
@@ -0,0 +1,365 @@
...
+    /**
+     * Transform an iterator of paths into a collection of resources
+     *
+     * @param \Iterator $iterator
+     * @return FilesystemResourceCollection
+     */
+    private function iteratorToCollection(\Iterator $iterator)
+    {
+        $paths = iterator_to_array($iterator);
+        $resources = array_values($this->store->getMultiple($paths));
```
webmozart: As far as I understand, you have a collection of paths here that you first need to turn into resources, no?

src/OptimizedPathMappingRepository.php
webmozart: Yes please :)

src/OptimizedPathMappingRepository.php
tgalopin: I have an iterator of path yes. Why?

src/OptimizedPathMappingRepository.php
```diff
@@ -0,0 +1,365 @@
...
+     */
+    private function getChildIterator(FilesystemResource $resource)
+    {
+        $staticPrefix = rtrim($resource->getPath(), '/').'/';
+        $regExp = '~^'.preg_quote($staticPrefix, '~').'[^/]+$~';
+
+        return new RegexFilterIterator(
+            $regExp,
+            $staticPrefix,
+            new \ArrayIterator($this->store->keys())
```
webmozart: I usually import all classes, even global ones. PHPStorm does this automatically, if you use it.

tgalopin: I use it, but probably did not configured it well. I'll change that :)

src/OptimizedPathMappingRepository.php
webmozart: You pass the paths to FilesystemResourceCollection, which expects an array of resources. Does that work?

src/OptimizedPathMappingRepository.php
webmozart: It's probably a good idea to write a test for that too

src/OptimizedPathMappingRepository.php
```diff
@@ -0,0 +1,365 @@
...
+        Assert::stringNotEmpty($path, 'The path must be a non-empty string. Got: %s');
+        Assert::startsWith($path, '/', 'The path %s is not absolute.');
+
+        $path = Path::canonicalize($path);
+
+        if (! $this->store->exists($path)) {
+            throw ResourceNotFoundException::forPath($path);
+        }
+
+        return $this->store->get($path);
```
tgalopin: Need also to attach the resource to the repo on get() (for serialization support).

src/OptimizedPathMappingRepository.php
```diff
@@ -0,0 +1,365 @@
...
+ * @since  1.0
+ * @author Bernhard Schussek <bschussek@gmail.com>
+ * @author Titouan Galopin <galopintitouan@gmail.com>
+ */
+class OptimizedPathMappingRepository implements EditableRepository
+{
+    /**
+     * @var KeyValueStore
+     */
+    private $store;
```
tgalopin: Let's try to store filesystem paths instead of resources in the store.

tgalopin:
I fixed all your notes @webmozart.

I'm working on the benchmarks to see if storing paths would be a better idea.

tgalopin:
Okay, I did the benchmarks, and here are the results. Please don't hesitate to tell me if you see a problem in the process I'll explain here.

**1. Raw data**

I had to generate a bit of raw data to benchmark properly. Here is the script doing so:

``` php
<?php

require 'src/vendor/autoload.php';

/*
 * Resources
 */
$store = new \Webmozart\KeyValueStore\ArrayStore();
$repository = new \Puli\Repository\OptimizedPathMappingRepository($store);

for ($i = 0; $i < 50; $i++) {
    $repository->add('/tgalopin/key' . $i, new \Puli\Repository\Resource\DirectoryResource(__DIR__ . '/fixtures'));
}

file_put_contents('serialized-resource.data', serialize($store));


/*
 * Paths
 */
$store = new \Webmozart\KeyValueStore\ArrayStore();
$repository = new \Puli\Repository\OptimizedPathMappingRepositoryPath($store);

for ($i = 0; $i < 50; $i++) {
    $repository->add('/tgalopin/key' . $i, new \Puli\Repository\Resource\DirectoryResource(__DIR__ . '/fixtures'));
}

file_put_contents('serialized-path.data', serialize($store));
```

This script will generate a first serialized file containing a store of resources, and a second one containing a store of paths.

**2. Benchmark process**

*2. a. Resources*

Repository: https://gist.github.com/tgalopin/9d0ce833d9bfcbdd2207

Here is the script:

``` php
<?php

require 'src/vendor/autoload.php';

$start = microtime(true);

for ($i = 0; $i < 1000; $i++) {
    $store = unserialize(file_get_contents('serialized-resource.data'));
    $repository = new \Puli\Repository\OptimizedPathMappingRepository($store);

    for ($j = 0; $j < 50; $j++) {
        $root = $repository->get('/tgalopin/key' . $j);
        $rootFile = $repository->get('/tgalopin/key' . $j . '/rootfile.txt');
        $sub = $repository->get('/tgalopin/key' . $j . '/sub');
        $subFile = $repository->get('/tgalopin/key' . $j . '/sub/subfile.txt');

        unset($root, $rootFile, $sub, $subFile);
    }
}

echo (microtime(true) - $start) . "ms per iteration\n";
```

*2. b. Paths*

Repository: https://gist.github.com/tgalopin/fa48225c87abcff29764

Here is the script:

``` php
<?php

require 'src/vendor/autoload.php';

$start = microtime(true);

for ($i = 0; $i < 1000; $i++) {
    $store = unserialize(file_get_contents('serialized-path.data'));
    $repository = new \Puli\Repository\OptimizedPathMappingRepositoryPath($store);

    for ($j = 0; $j < 50; $j++) {
        $root = $repository->get('/tgalopin/key' . $j);
        $rootFile = $repository->get('/tgalopin/key' . $j . '/rootfile.txt');
        $sub = $repository->get('/tgalopin/key' . $j . '/sub');
        $subFile = $repository->get('/tgalopin/key' . $j . '/sub/subfile.txt');

        unset($root, $rootFile, $sub, $subFile);
    }
}

echo (microtime(true) - $start) . "ms per iteration\n";
```

**3. Results**

I ran both the scripts ten times (`for i in {1..10}; do php benchmark.php; done`) and got these results:

*3. a. Resources*

``` text
3.7852931022644ms per iteration
3.8438930511475ms per iteration
3.8876490592957ms per iteration
3.765016078949ms per iteration
3.8802471160889ms per iteration
3.9678199291229ms per iteration
3.7913291454315ms per iteration
3.8928418159485ms per iteration
3.9371881484985ms per iteration
3.7685868740082ms per iteration
```

*3. b. Paths*

``` text
4.9362621307373ms per iteration
5.1432459354401ms per iteration
5.0016570091248ms per iteration
5.1690270900726ms per iteration
4.8666050434113ms per iteration
4.9411540031433ms per iteration
4.8729741573334ms per iteration
4.994236946106ms per iteration
4.9226040840149ms per iteration
4.9880139827728ms per iteration
```

Resources seem to be faster. However, I'm not sure about my process and I may have introduce some comparison problems, so don't hesitate to search for them.

For the moment, I suggest we keep direct resources storage.

tgalopin:
I implemented the storage of paths in the KeyValue store instead of resources, according to benchmarks. I also improved the removing of resources by not accessing the filesystem in the process.

I think this PR is ready to be merged and I'll work on the classic PathMappingRepository now.If you have suggestions about the current code, please tell me, I'll be pleased to improve it!

src/OptimizedPathMappingRepository.php
```diff
@@ -0,0 +1,464 @@
...
+        return $resource->getFilesystemPath();
+    }
+
+    /**
+     * Create a resource from a given string of the store
+     *
+     * @param string $serializedResource
+     * @return FilesystemResource
+     */
+    private function unserialize($serializedResource)
```
webmozart: I would rename this method to `createResource()` and the argument to `$filesystemPath`. Then it's more obvious what the method actually does.

src/OptimizedPathMappingRepository.php
```diff
@@ -0,0 +1,464 @@
...
+        }
+    }
+
+    /**
+     * Create a string representing a given resource for the store
+     *
+     * @param FilesystemResource $resource
+     * @return string
+     */
+    private function serialize(FilesystemResource $resource)
```
webmozart: I don't think this method has any benefit. I'd use $resource->getFilesystemPath() right away, then it's obvious what is happening. At the moment, you could believe that the method serializes the resource.

src/OptimizedPathMappingRepository.php
webmozart: This method looks rather slow. I think we should try two approaches:

1. Add sorting directly to the key-value-stores that support it (see webmozart/key-value-store#8)
2. If a KVS does not support sorting, maintain a separate entry (like `_key_order`) that stores all keys in order. It should be much faster to sort one entry than sorting the complete store.

src/OptimizedPathMappingRepository.php
```diff
@@ -0,0 +1,464 @@
...
+        }
+    }
+
+    /**
+     * Transform an iterator of paths into a collection of resources
+     *
+     * @param \Iterator $iterator
+     * @return FilesystemResourceCollection
+     */
+    private function iteratorToCollection(\Iterator $iterator)
```
webmozart: Puli always imports global classes:

```php
use Iterator;

private function iteratorToCollection(Iterator $iterator)
```

src/OptimizedPathMappingRepository.php
```diff
@@ -0,0 +1,464 @@
...
+    /**
+     * Count the number of elements in the store
+     */
+    private function createRoot()
+    {
+        if ($this->store->exists('/')) {
+            return;
+        }
+
+        $this->store->set('/', null);
```
webmozart: Why do you set this to `null` and not a `GenericResource`? This way, it's impossible to do `$repo->get('/')->getChild('foo')`, no?

src/Resource/GenericFilesystemResource.php
```diff
@@ -0,0 +1,257 @@
...
+use Puli\Repository\Resource\Collection\ArrayResourceCollection;
+
+/**
+ * A generic filesystem resource.
+ *
+ * @since  1.0
+ * @author Bernhard Schussek <bschussek@gmail.com>
+ * @author Titouan Galopin <galopintitouan@gmail.com>
+ */
+class GenericFilesystemResource implements FilesystemResource
```
webmozart: I don't think we need this class. Let's use a `GenericResource` and a normal `ArrayResourceCollection` instead.

tests/Fixtures/path_mapping/file0
```diff
@@ -0,0 +1,2 @@
+LINE 1
```
webmozart: Might be a stupid question: Can the tests be written to use the current fixtures?

If not, that's not a big deal. But I prefer to have as little fixtures as possible, as it's hard to keep track which fixtures are used where.

tgalopin: Should be possible, I just created these fixtures automatically. I'll see :) .

src/OptimizedPathMappingRepository.php
tgalopin: Okay :) .

src/OptimizedPathMappingRepository.php
tgalopin: I thought a serialize/unserialize system could be more explicit than something in each method.But no problem, I'll do that

src/OptimizedPathMappingRepository.php
tgalopin: Yup, PhpStorm misconfiguredd :)

src/OptimizedPathMappingRepository.php
tgalopin: No, as I store paths in the store, I create a GenericResource on `unserialize()` if the path is null.

src/Resource/GenericFilesystemResource.php
tgalopin: But then it means the repository supports other resources than just FilesystemResource ones doesn't it?

tests/OptimizedPathMappingRepositoryTest.php
```diff
@@ -0,0 +1,159 @@
...
+    }
+
+    protected function createReadRepository(EditableRepository $writeRepo)
+    {
+        return $writeRepo;
+    }
+
+    protected function createFile($path = null, $body = TestFilesystemFile::BODY)
+    {
+        $filesystemPath = __DIR__ . '/Fixtures/path_mapping/file' . self::$createdFiles;
```
webmozart: You should work in the system's temporary directory, not in the fixtures directory, otherwise we might commit files created by the test by accident. See FilesystemRepositoryTest::setUp() and tearDown() for inspiration.

src/OptimizedPathMappingRepository.php
webmozart: https://youtrack.jetbrains.com/issue/WI-26767

src/Resource/GenericFilesystemResource.php
webmozart: Yes and no. You can only pass FilesystemResources to add(), but the repository may return either GenericResource or FilesystemResource. If you work with a ResourceRepository (without knowing the implementation) and want to access the filesystem path, you need to do a `instanceof FilesystemResource` check anyway, so that's not a big problem.

webmozart:
Thanks a lot for this great PR @tgalopin! :)

src/Resource/GenericFilesystemResource.php
tgalopin: Ok, I see what you mean. I'll remove it then.

tests/OptimizedPathMappingRepositoryTest.php
tgalopin: I'll, I never thought of this :) .

src/OptimizedPathMappingRepository.php
tgalopin: :) Thanks

tests/OptimizedPathMappingRepositoryTest.php
tgalopin: Done :)

tgalopin:
I chaned various things accordingly to your notes. Do you have any other thing in mind to change?

src/OptimizedPathMappingRepository.php
```diff
@@ -0,0 +1,457 @@
...
+
+    /**
+     * Recursively creates a directory for a path.
+     *
+     * @param string $path A directory path.
+     * @return DirectoryResource The created resource
+     */
+    private function ensureDirectoryExists($path)
+    {
+        if (!$this->store->exists($path)) {
```
webmozart: Could you invert the condition and add an early return instead? (like in createRoot())

src/OptimizedPathMappingRepository.php
```diff
@@ -0,0 +1,457 @@
...
+            // Recursively initialize parent directories
+            if ($path !== '/') {
```
webmozart: Please use Yoda ordering (`if ('/' !== $path) {`)

src/OptimizedPathMappingRepository.php
webmozart: Ah sure, thanks for clarifying :)

webmozart:
Looks good now, thanks! Could you fix the remaining issues?

Do you want to work on sortable key-value stores next? (webmozart/key-value-store#8) Then we could improve the sorting used here.

tgalopin:
Done :) .

The sorting issue seems interesting, I'll give it a shot. Perhaps after the classical PathMapping repository?

webmozart:
Thanks!

PR #28, opened 2015-07-04
Fix build on Windows

WouterJ:
Fixes https://github.com/puli/issues/issues/65

* Some tests were comparing paths against PHP internal functions (like `readlink`). These still return the Windows variant of the path. That's why I introduced `assertPathsAreEqual` in some test cases.
* Relative symlinks aren't supported in PHP on Windows, marked tests as skipped on windows builds.
* The filesystem resources were storing the Windows variant of the filesystem path, meaning that comparing 2 objects (one created by a repository, one created in the test) resulted in problems. Now, the filesystem paths are also normalized.

### DIFF
--- a/tests/AbstractFilesystemRepositoryTest.php
+++ b/tests/AbstractFilesystemRepositoryTest.php
@@ -1,0 +1,28 @@
+<?php
+
+/*
+ * This file is part of the puli/repository package.
+ *
+ * (c) Bernhard Schussek <bschussek@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Puli\Repository\Tests;
+
+/**
+ * @since  1.0
+ * @author Bernhard Schussek <bschussek@gmail.com>
+ */
+abstract class AbstractFilesystemRepositoryTest extends AbstractEditableRepositoryTest
+{
+    protected function assertPathsAreEqual($expected, $actual)
+    {
+        $normalize = function ($path) {
+            return str_replace(DIRECTORY_SEPARATOR, '/', $path);
+        };
+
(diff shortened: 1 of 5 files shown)


tests/AbstractFilesystemRepositoryTest.php
```diff
@@ -0,0 +1,28 @@
...
+ */
+abstract class AbstractFilesystemRepositoryTest extends AbstractEditableRepositoryTest
+{
+    protected function assertPathsAreEqual($expected, $actual)
+    {
+        $normalize = function ($path) {
+            return str_replace(DIRECTORY_SEPARATOR, '/', $path);
+        };
+
+        $this->assertEquals($normalize($expected), $normalize($actual));
```
xabbuh: The tests originally used `assertSame()`.

WouterJ: there is no need for `assertSame`, as both are passed through `str_replace`, meaning they always have the same type.

tests/Resource/AbstractFilesystemResourceTest.php
```diff
@@ -97,6 +97,15 @@ public function testSerializeKeepsFilesystemPath()
...
+        $this->assertPathsAreEqual($filesystemPath, $deserialized->getFilesystemPath());
+    }
+
+    protected function assertPathsAreEqual($expected, $actual)
+    {
+        $normalize = function ($path) {
+            return str_replace(DIRECTORY_SEPARATOR, '/', $path);
+        };
+
+        $this->assertEquals($normalize($expected), $normalize($actual));
```
webmozart: Isn't this a duplicate of the method in the base class?

webmozart:
Thanks @WouterJ!

PR #29, opened 2015-07-09
[READY] Implement PathMappingRepository (development version of OptimizedPathMappingRepository)

tgalopin:
Following the OptimizedPathMapping repository (puli/repository#27), I implemented a development version of this repository, as proposed by @webmozart in puli/issues#80. This pull-request is a place to debate over the code.

I did three main things:

**1. Create the repository**

https://github.com/tgalopin/repository/blob/dev-path-mapping/src/PathMappingRepository.php

The repository follows these rules:
- When a resource is added, the association `repositoryPath => filesystemPath` is stored in the KeyValueStore, and the parents of the repositoryPath are created as virtual resources ;
-  When a resource is fetched (using `get()` or other methods), its repositoryPath is *resolved* to a filesystemPath: the repository find the list of virtual resources matching the searched resource repositoryPath and try to find a real directory / file corresponding to the resource in each of these virtual resources ;

This technique is much slower than the optimized version (obviously :) ) but it's a really useful tool for development as the structure inside virtual resources can change without having to "recreate" the puli repository.

**2. Add a build structure system in tests**

https://github.com/tgalopin/repository/blob/dev-path-mapping/tests/AbstractRepositoryTest.php#L62

This repository was not possible to test using the exact same tests as before as when we are requesting a resources, it can be either a virtual or a real one, and tests have to check for both at the same time.

I had to create a real valid structure on filesystem according to the tests: I couldn't without altering a bit the existing tests.

I added a `buildStructure` method, called on each built resources (for instance https://github.com/tgalopin/repository/blob/dev-path-mapping/tests/AbstractRepositoryTest.php#L102) that by default return the root resource to be compatible with other repositories tests.

But in the PathMappingRepository tests, this method create the structure (https://github.com/tgalopin/repository/blob/dev-path-mapping/tests/PathMappingRepositoryTest.php#L113) for tests to work properly.

**3. Create the specific tests for the PathMappingRepository**

Once the built structure ready, I fixed the repository to pass the tests (https://github.com/tgalopin/repository/blob/dev-path-mapping/tests/PathMappingRepositoryTest.php).

I'm very open to modifications, expecially on how `find()` and `contains()` work as I'm not sure it's the best algorithm possible.

### DIFF
--- a/src/AbstractPathMappingRepository.php
+++ b/src/AbstractPathMappingRepository.php
@@ -1,0 +1,170 @@
+<?php
+
+/*
+ * This file is part of the puli/repository package.
+ *
+ * (c) Bernhard Schussek <bschussek@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Puli\Repository;
+
+use Countable;
+use Puli\Repository\Api\EditableRepository;
+use Puli\Repository\Resource\DirectoryResource;
+use Puli\Repository\Resource\FileResource;
+use Puli\Repository\Resource\GenericResource;
+use Webmozart\KeyValueStore\Api\KeyValueStore;
+use Webmozart\PathUtil\Path;
+
+/**
+ * Abstract base for Path mapping repositories.
+ *
+ * @since  1.0
+ *
+ * @author Bernhard Schussek <bschussek@gmail.com>
+ * @author Titouan Galopin <galopintitouan@gmail.com>
+ */
+abstract class AbstractPathMappingRepository implements EditableRepository
+{
+    /**
+     * @var KeyValueStore
+     */
+    protected $store;
+
+    /**
+     * Creates a new repository.
+     *
+     * @param KeyValueStore $store The store of all the paths.
+     */
+    public function __construct(KeyValueStore $store)
+    {
+        $this->store = $store;
+
+        $this->createRoot();
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function clear()
+    {
+        // Subtract root
+        $removed = $this->countStore() - 1;
+
+        $this->store->clear();
+        $this->createRoot();
+
+        return $removed;
+    }
+
+    /**
+     * Recursively creates a directory for a path.
+     *
+     * @param string $path A directory path.
+     */
+    protected function ensureDirectoryExists($path)
+    {
+        if ($this->store->exists($path)) {
+            return;
+        }
+
+        // Recursively initialize parent directories
+        if ('/' !== $path) {
+            $this->ensureDirectoryExists(Path::getDirectory($path));
+        }
+
+        $this->store->set($path, null);
+    }
+
+    /**
+     * Create the repository root.
+     */
+    protected function createRoot()
+    {
+        if ($this->store->exists('/')) {
+            return;
+        }
+
+        $this->store->set('/', null);
+    }
+
+    /**
+     * Count the number of elements in the store.
+     *
+     * @return int
+     */
+    protected function countStore()
+    {
+        if ($this->store instanceof Countable) {
+            return count($this->store);
+        }
+
+        return count($this->store->keys());
+    }
+
+    /**
+     * Sort the store by keys.
+     */
+    protected function sortStore()
+    {
+        $resources = $this->store->getMultiple($this->store->keys());
+
+        ksort($resources);
+
+        $this->store->clear();
+
+        foreach ($resources as $path => $resource) {
+            $this->store->set($path, $resource);
+        }
+    }
+
+    /**
+     * Create a resource using its filesystem path.
+     *
+     * If the filesystem path is a directory, a DirectoryResource will be created.
+     * If the filesystem path is a file, a FileResource will be created.
+     * If the filesystem does not exists, a GenericResource will be created.
+     *
+     * @param string $filesystemPath The filesystem path.
+     *
+     * @return GenericResource|DirectoryResource|FileResource The created resource.
+     */
+    protected function createResource($filesystemPath)
+    {
+        if ($filesystemPath === null) {
+            return new GenericResource();
+        }
+
+        if (is_dir($filesystemPath)) {
+            return new DirectoryResource($filesystemPath);
+        } elseif (is_file($filesystemPath)) {
+            return new FileResource($filesystemPath);
+        }
+
+        return new GenericResource();
+    }
+
+    /**
+     * Create a collection of resources using a list of filesystem paths.
+     *
+     * @param array $filesystemPaths
+     *
+     * @return array
+     */
+    protected function createResources($filesystemPaths)
+    {
+        $resources = array();
+
+        foreach ($filesystemPaths as $path => $filesystemPath) {
+            $child = $this->createResource($filesystemPath);
+            $child->attachTo($this, $path);
+
+            $resources[] = $child;
+        }
+
+        return $resources;
+    }
+}
--- a/src/AbstractPathMappingRepository.php
+++ b/src/AbstractPathMappingRepository.php
@@ -12,7 +12,6 @@
 namespace Puli\Repository;
 
 use Countable;
-use Puli\Repository\Api\EditableRepository;
 use Puli\Repository\Resource\DirectoryResource;
 use Puli\Repository\Resource\FileResource;
 use Puli\Repository\Resource\GenericResource;
@@ -27,7 +26,7 @@ use Webmozart\PathUtil\Path;
  * @author Bernhard Schussek <bschussek@gmail.com>
  * @author Titouan Galopin <galopintitouan@gmail.com>
  */
-abstract class AbstractPathMappingRepository implements EditableRepository
+abstract class AbstractPathMappingRepository extends AbstractRepository
 {
     /**
      * @var KeyValueStore
--- a/src/AbstractPathMappingRepository.php
+++ b/src/AbstractPathMappingRepository.php
@@ -12,9 +12,13 @@
 namespace Puli\Repository;
 
 use Countable;
+use Puli\Repository\Api\Resource\FilesystemResource;
+use Puli\Repository\Api\ResourceCollection;
+use Puli\Repository\Api\UnsupportedResourceException;
 use Puli\Repository\Resource\DirectoryResource;
 use Puli\Repository\Resource\FileResource;
 use Puli\Repository\Resource\GenericResource;
+use RuntimeException;
 use Webmozart\KeyValueStore\Api\KeyValueStore;
 use Webmozart\PathUtil\Path;
 
@@ -43,6 +47,47 @@ abstract class AbstractPathMappingRepository extends AbstractRepository
         $this->store = $store;
 
         $this->createRoot();
+    }
+
+    /**
+     * Add the resource (internal method after checks of add()).
+     *
+     * @param string             $path
+     * @param FilesystemResource $resource
+     */
+    abstract protected function addResource($path, FilesystemResource $resource);
+
+    /**
+     * {@inheritdoc}
+     */
+    public function add($path, $resource)
+    {
+        $path = $this->sanitizePath($path);
+
+        if ($resource instanceof ResourceCollection) {
+            $this->ensureDirectoryExists($path);
+
+            foreach ($resource as $child) {
+                $this->addResource($path.'/'.$child->getName(), $child);
+            }
+
+            $this->sortStore();
+
+            return;
+        }
+
+        if ($resource instanceof FilesystemResource) {
+            $this->ensureDirectoryExists(Path::getDirectory($path));
+            $this->addResource($path, $resource);
+            $this->sortStore();
+
+            return;
+        }
+
+        throw new UnsupportedResourceException(sprintf(
+            'The passed resource must be a FilesystemResource or a ResourceCollection. Got: %s',
+            is_object($resource) ? get_class($resource) : gettype($resource)
+        ));
     }
 
     /**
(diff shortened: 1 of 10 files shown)


tgalopin:
I think this is ready. Could you review it @webmozart :) ?

src/PathMappingRepository.php
```diff
@@ -0,0 +1,547 @@
...
+        return $this->getChildren($this->get($path))->count() !== 0;
+    }
+
+
+    /**
+     * Find a resource by its path.
+     *
+     * @param string $path The path to resolve.
+     *
+     * @return \Puli\Repository\Api\Resource\Resource|null  The resource or null if
```
webmozart:
```
@return Resource|null
```

src/PathMappingRepository.php
```diff
@@ -0,0 +1,547 @@
...
+         *          file or directory on the filesystem and if we do find one,
+         *          we stop
+         */
+        $storePaths = array_reverse($this->store->keys());
+
+        // Create regex
+        $regExpressions = array();
+
+        foreach ($storePaths as $storePath) {
+            $prefix = rtrim($storePath, '/') . '/';
```
webmozart: Please remove the wrapping spaces around the `.` (everywhere)

src/PathMappingRepository.php
```diff
@@ -0,0 +1,547 @@
...
+            $regExpressions[] = '~^' . preg_quote($prefix, '~') . '~';
+        }
+
+        // Find potential parents
+        $resolved = null;
+
+        foreach ($regExpressions as $key => $regExpression) {
```
webmozart: You don't really need to use regular expressions here (regexes are slow). I'd do something like:

```php
$basePaths = array_reverse($this->store->keys());

foreach ($basePaths as $basePath) {
    if (!Path::isBasePath($basePath, $path)) {
        continue;
    }

    $basePath = rtrim($basePath, '/').'/';
    $filesystemBasePath = rtrim($this->store->get($basePath), '/').'/';
    $filesystemPath = substr_replace($path, $filesystemBasePath, 0, strlen($basePath));

    // ...
}
```

Additionally, you probably want to do:

```php
if (!file_exists($filesystemPath)) {
    continue;
}
```

in the loop and throw a ResourceNotFoundException at the end of the method. Currently, if multiple mappings are base paths for the searched path and the path is not found in the *first* mapping, this method fails even if it exists in the second mapping. You should add a test for that.

src/PathMappingRepository.php
tgalopin: The problem with this is that PHPStorm and other IDE think it means the `resource` internal type. Should I change it anyway?

src/PathMappingRepository.php
```diff
@@ -0,0 +1,547 @@
...
+                $filesystemRoot = rtrim($this->store->get($storePaths[$key]), '/') . '/';
+                $filesystemPath = preg_replace($regExpression, $filesystemRoot, $path);
+
+                $resource = $this->createResource($filesystemPath);
+
+                // The resource is resolved
+                if ($resource instanceof FilesystemResource) {
+                    $resolved = $resource;
+                    $resolved->attachTo($this, $path);
+                    break;
```
webmozart: You can directly return here.

What happens when accessing a virtual resource? It should be possible to access them too, no?

src/PathMappingRepository.php
webmozart: Yes, that's the problem of PHPStorm IMO.

src/PathMappingRepository.php
tgalopin: The virtual resources don't need to be resolved, they are directly in the store.

Why do you think I should directly return? I don't have to attach the resource you mean?

src/PathMappingRepository.php
webmozart: Apart from that, Resource may have to be renamed anyway if it really becomes a reserved keyword in PHP7 :P

src/PathMappingRepository.php
tgalopin: > Currently, if multiple mappings are base paths for the searched path and the path is not found
> in the first mapping, this method fails even if it exists in the second mapping.

The `createResource()` and `$resource instanceof FilesystemResource` part is all about finding if the real filesystem file exists. It will stop the loop only if the filesystem resource does exists.

But you are right, I'll write a test for this.

src/PathMappingRepository.php
tgalopin: Okay, no problem.

src/PathMappingRepository.php
tgalopin: Well, it's a big problem but it's different one :D . I'll make the change.

webmozart:
Could you rebase this on 1.0 please?

tgalopin:
I have no idea how to do that :D .

webmozart:
```
git remote add puli git@github.com:puli/repository.git
git fetch puli
git checkout your_branch
git rebase puli/1.0
# fix merge errors if any
git push
```

tgalopin:
I'm not sure of the rebase ... Is it ok?

webmozart:
-> IRC

tgalopin:
I improved the code, but I think I can't really fix https://scrutinizer-ci.com/g/puli/repository/inspections/59ac0d0d-c47a-4770-bd64-dc22429283ad/code-structure/operation/Puli%5CRepository%5CPathMappingRepository%3A%3Aget as it's not a real code duplication. What do you think?

webmozart:
What about adding another AbstractFilesystemRepository as that contains all the common methods for all file system based repos? That would solve the problem, no?

tgalopin:
I think so but it requires a bit more reflexion than just copy-past as it's not the real same methods. I'll think about it.

webmozart:
Ok, thanks :)

tgalopin:
I don't understand the message of Scrunitizer:

> Consider making the return type a bit more specific; maybe use NoType.

It's on a `return int`, I don't see the problem. Do you?

tgalopin:
I squashed my 23 (\o/) commits in 5 with clearer messages for the sanity of your history :) .

src/PathMappingRepository.php
```diff
@@ -0,0 +1,382 @@
...
+ */
+class PathMappingRepository extends AbstractPathMappingRepository implements EditableRepository
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function get($path)
+    {
+        $path = $this->sanitizePath($path);
+        $resource = $this->resolveResource($path);
```
webmozart: What happens for virtual resources? For example, if I do:

```php
$repo->add('/foo/bar', new DirectoryResource(...));
```

then I should get:

```php
$repo->get('/foo')->get('bar') => DirectoryResource
```

But as far as I see, virtual resources such as `/foo` can't be accessed with this implementation, can they?

src/PathMappingRepository.php
```diff
@@ -0,0 +1,382 @@
...
+     *
+     * @return Resource[] The children.
+     */
+    private function getFilesystemResourceChildren(FilesystemResource $resource)
+    {
+        if (!is_dir($resource->getFilesystemPath())) {
+            return array();
+        }
+
+        $iterator = new \RecursiveDirectoryIterator(
```
webmozart: Please import global classes.

src/PathMappingRepository.php
```diff
@@ -0,0 +1,382 @@
...
+     * @param Resource $resource The resource.
+     *
+     * @return ArrayResourceCollection The children.
+     */
+    private function getChildrenRecursive(Resource $resource)
+    {
+        $resources = new ArrayResourceCollection();
+
+        foreach ($this->getChildren($resource) as $child) {
+            $resources->merge($this->getChildrenRecursive($child)->toArray());
```
webmozart: I think this method should be optimized. In case of FilesystemResources, you should immediately do a recursive iteration instead of subsequent calls to `getChildren()`.

src/PathMappingRepository.php
```diff
@@ -0,0 +1,382 @@
...
+         * Otherwise, we need to search for paths matching the glob:
+         *      1.  find the root resource of all search results by
+         *          resolving the glob base path
+         *      2.  find all the children of this root resource and
+         *          try to match their path to the query
+         */
+        $basePath = Glob::getBasePath($query);
+        $baseResource = $this->resolveResource($basePath);
+
+        if ($baseResource !== null) {
```
webmozart: Please use YODA expressions: `null !== $baseResource`

Hopefully this can be done by the CS fixer soon.

webmozart: Also, I'd turn this into an early return:

```php
if (null === $baseResource) {
    return $resources;
}
```

In general, early returns should be preferred - if possible - as they reduce the nesting level of a function and make the control flow better visible.

src/PathMappingRepository.php
```diff
@@ -0,0 +1,382 @@
...
+            $children = $this->getChildrenRecursive($baseResource);
+
+            $i = 0;
+            $found = false;
+
+            while ($i < $children->count() && !($found && $singleResult)) {
```
webmozart: I'd use a `foreach` with a `break` here - again, clearer what the loop does and when it quits.

tests/PathMappingRepositoryTest.php
```diff
@@ -0,0 +1,405 @@
...
+    {
+        return new TestFilesystemFile($path, $body);
+    }
+
+    protected function createDirectory($path = null, array $children = array())
+    {
+        return new TestFilesystemDirectory($path, $children);
+    }
+
+    protected function buildStructure(Resource $root)
```
webmozart: Can this method be merged into `createDirectory()` and `createFile()` or does it really need to be called explicitly in the tests?

webmozart:
Thanks @tgalopin! This is starting to look really good.

src/PathMappingRepository.php
tgalopin: Why couldn't they? `resolveResource` do check first if the resource directly exists in the store (first condition) so as `/foo` will be present there, it will work. I'm pretty sure the tests wouldn't pass if  `$repo->get('/foo')->get('bar') => DirectoryResource` wasn't possible.

src/PathMappingRepository.php
tgalopin: I changed it because of ScrunitizerCI: using a `while` avoid a condition and is a bit more clearer than foreach about when it stops. Should i ignore the Scrunitizer notice?

src/PathMappingRepository.php
tgalopin: I'll do that.

src/PathMappingRepository.php
tgalopin: Yep, true. I'll do that.

src/PathMappingRepository.php
tgalopin: Sorry :D . Isn't it possible to configure StyleCI for that?

tests/PathMappingRepositoryTest.php
tgalopin: I tried to merge it but I can't really know how to build a structure from the inside of `createDirectory` for instance. I mean, for instance, in the following coded, I'm not able to know that the first createDirectory should build the file structure where the inside one should not:

``` php
$this->createDirectory('', array(
    $this->createDirectory('/sub1', array(
        $this->createFile('/file1', 'original 1'),
    )),
    $this->createDirectory('/sub2', array(
        $this->createFile('/file2', 'original 2'),
    )),
))
```

src/PathMappingRepository.php
webmozart: Not at the moment. See FriendsOfPHP/PHP-CS-Fixer#1309

src/PathMappingRepository.php
webmozart: Hm. Which of the two did you find more readable?

src/AbstractPathMappingRepository.php
```diff
@@ -0,0 +1,169 @@
...
+     * If the filesystem path is a file, a FileResource will be created.
+     * If the filesystem does not exists, a GenericResource will be created.
+     *
+     * @param string $filesystemPath The filesystem path.
+     *
+     * @return GenericResource|DirectoryResource|FileResource The created resource.
+     */
+    protected function createResource($filesystemPath)
+    {
+        if ($filesystemPath === null) {
```
webmozart: YODA

webmozart: In fact, I think this method is a bit confusing. The `null` case should be handled outside this method IMO, since there's only one call that actually needs it - in `get()`.

Why does it return GenericResources if a file is not found?

src/PathMappingRepository.php
tgalopin: I prefer the foreach. but Scrunitizer is not very open to discussion :D .

src/PathMappingRepository.php
tgalopin: I just added a test for that:

https://github.com/puli/repository/blob/cffc57a01bfb2fda03a6f1dec64e031aa17f7581/tests/PathMappingRepositoryTest.php#L182

src/PathMappingRepository.php
webmozart: Then change it to the foreach. Scrutinizer can be configured. ;)

src/AbstractPathMappingRepository.php
tgalopin: The GenericResource represents a virtual resource. That's why it's returned when there is no file corresponding. But I agree, the null is not really interesting here anymore (it was when the code was directly in the method).

webmozart: I understand that. However, for virtual resources, the filesystem path should always be null, hence this method shouldn't even be called, no?

tgalopin: Yep, I'll refactor that to improve the logic.

tgalopin:
Okay, I think it's ready. I'll work on the integration of sortable and countable stores after this PR.

src/AbstractPathMappingRepository.php
```diff
@@ -0,0 +1,263 @@
...
+
+    /**
+     * Check if the given filesystem path exists and can be used for a
+     * filesystem resource.
+     *
+     * @param string $path The path.
+     *
+     * @return bool If the filesystem path exists
+     */
+    protected function isFilesystemPath($path)
```
webmozart: You can simply use `file_exists()` instead of this method. :)

tgalopin:
Compatible PHP7 \o/ :) .

webmozart:
Merged :) Thanks @tgalopin!

tgalopin:
I don't know the process, perhaps is it automatic, but for the moment the API doc is not updated: http://api.puli.io/latest/ (it does not contain the new repos :) ).

webmozart:
Updated now :) The docs are updated by checking out https://github.com/puli/api.puli.io, running `make` and pushing.

PR #32, opened 2015-07-28
[READY] Implement usage of sortable stores in path mapping repositories

tgalopin:
Implement the Sortable stores (https://github.com/webmozart/key-value-store/pull/9) in PathMapping repositories.


tgalopin:
It seems there is a problem with the StyleCI parsing, look like a bug of the platform: https://styleci.io/analyses/qx7jAX

webmozart:
Thanks! :)

PR #34, opened 2015-07-28
Fixed typos

GrahamCampbell:


GrahamCampbell:
This should fix all your issues with StyleCI.

webmozart:
@GrahamCampbell Cool thanks! :) What was the problem?

GrahamCampbell:
That missing brace was causing the fixed file to have a syntax error.

webmozart:
Ah! Is there (in general) any way to debug such problems without needing your assistance?

GrahamCampbell:
I manually disabled our linting process to highlight that issue. I'm about to ship that feature to production though, so soon you'll be able to just add `linting: false` to your config to highlight the exact issue.

webmozart:
Ok :) Will there be a way to manually trigger a scan with a changed config? I don't really want to push two commits (`linting: false`, `linting: true`) every time there is a problem.

GrahamCampbell:
> Ok :) Will there be a way to manually trigger a scan with a changed config? I don't really want to push two commits (linting: false, linting: true) every time there is a problem.

Absolutely. You can actually modify the config from in your browser you you don't have to push any commits. Just update the config in browser, then hit the analyze now button.

webmozart:
Ah! I hadn't seen that, cool :) Thanks a lot!

GrahamCampbell:
We shipped that feature yesterday with our redesign. :)

PR #36, opened 2015-08-04
Fix phpdoc type

mnapoli:
Was previously using `\Resource` instead of `Puli\Repository\Api\Resource\Resource`.

This is quite confusing when using the `ResourceRepository` as there is no auto-completion after `get()`.

However be aware: when testing locally, it seems PhpStorm has a bug and ignores the `use …`. As a consequence, it still believes the return type to be `\Resource`. I am on the latest EAP so I hope it's just a bug with the EAP.

### DIFF
--- a/src/Api/EditableRepository.php
+++ b/src/Api/EditableRepository.php
@@ -12,6 +12,7 @@
 namespace Puli\Repository\Api;
 
 use InvalidArgumentException;
+use Puli\Repository\Api\Resource\Resource;
 
 /**
  * A repository that supports the addition and removal of resources.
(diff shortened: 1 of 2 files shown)


webmozart:
Hi @mnapoli, thank you for this PR! Don't you think this is something that should be fixed in PHPStorm instead?

mnapoli:
Sorry didn't explain it right ;)

- current status: invalid phpdoc, not recognized by PhpStorm, the type resolves to `\Resource` (wrong class)
- with the PR: correct phpdoc, but still not recognized by PhpStorm (at least the last EAP)

I hope it will be fixed somehow on the PhpStorm side, but at least you'll have a valid phpdoc (e.g. here it's currently wrong: http://api.puli.io/latest/class-Puli.Repository.Api.ResourceRepository.html#_get).

webmozart:
Oops. I should have actually looked at the diff :) Thanks!

PR #37, opened 2015-08-06
Fix Windows symlink time information

WedgeSama:
On Windows, symlink does not work like on Unix, the OS think it is a real file like another. So for time operations, it only see changes on the symlink file and not the original one. That mean fileXtime functions get info about the symlink file too.

So if you make some changes in the original file, Puli will not seen anything if this file is register in the repository as a symlink. It work as espected for files in a symlink directory.

### DIFF
--- a/src/Resource/Metadata/FilesystemMetadata.php
+++ b/src/Resource/Metadata/FilesystemMetadata.php
@@ -17,7 +17,6 @@ use Puli\Repository\Api\Resource\ResourceMetadata;
  * Metadata about a file on the filesystem.
  *
  * @since  1.0
- * 
  * @author Bernhard Schussek <bschussek@gmail.com>
  */
 class FilesystemMetadata extends ResourceMetadata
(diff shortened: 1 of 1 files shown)


src/Resource/Metadata/FilesystemMetadata.php
```diff
@@ -30,14 +30,32 @@ public function __construct($filesystemPath)
...
 
     /**
+     * On Windows, fileXtime functions see only changes
+     * on the symlink file and not the original one.
+     *
+     * @param string $path
+     *
+     * @return string
+     */
+    private function fixWindowsPath($path)
```
webmozart: Could you move this method after the public methods please?

src/Resource/Metadata/FilesystemMetadata.php
```diff
@@ -17,7 +17,7 @@
  * Metadata about a file on the filesystem.
  *
  * @since  1.0
- *
+ *
```
webmozart: You seem to have added a space by mistake

webmozart:
Thanks a lot @WedgeSama! :) Looks good!

WedgeSama:
Done. I added the space cause StyleCI cry on this line but I remove it like you ask.

webmozart:
Thanks, merged! :)

GrahamCampbell:
> Done. I added the space cause StyleCI cry on this line but I remove it like you ask.

That's not what you removed...

GrahamCampbell:
You removed the entire line, instead of the single whitespace character on the end.

webmozart:
@GrahamCampbell I fixed the StyleCI issue in a separate commit.

WedgeSama:
Oh, that just that. I did not seen the whitespace.

GrahamCampbell:
Note that in future, you can just download and apply the diff styleci gives you.

WedgeSama:
Thank for the tips.

PR #39, opened 2015-08-18
[WIP] Support deletion in PathMappingRepository

tgalopin:
Pull request for https://github.com/puli/issues/issues/105

A question I'm asking myself is what to do if a resource is not found for deletion. For the moment, if the `remove()` method does not found any path to remove, it returns 0, which seems fine. However, as it's a specific context (PathMapping repository does only supports remove on mapped directories), we might want to throw a BadMethodException on invalid remove call.

What do you think?


tgalopin:
Oh and this pull request just port the `remove` method from the optimized repo into the abstract path mapping repo as the method would be the same in both of the repositories (optimized and dev).

webmozart:
Intuitively I'd say that `remove()` should fail if a path is removed that exists, but is not a path mapping. If a path does not exist, the method should simply return 0. What do you think?

tgalopin:
I think it's a good idea but we have to keep in mind that `remove` supports glob. What happens if we try to remove both valid and invalid paths? If we throw an exception at some point, we have either to remove all the removable paths or not remove any of them, but not stop in the middle.

I think as the PathMapping repository will be used mainly in development context, we definitely should throw an exception to warn the developer of the misuse. I'm just trying to find the best moment to do so.

tgalopin:
I rebase my work on https://github.com/puli/repository/pull/40 as it changes quite a lot of things.

tgalopin:
This will be fixed in #46 so I close here.

PR #40, opened 2015-08-19
[WIP] Fix #104 - Override does not work with repository type path-mapping

tgalopin:
Work to fix https://github.com/puli/issues/issues/104.

The tests were passing as the merge was made by the filesystem, not by the PathMapping repository. I fixed it to show the problem and I'm working on making the tests pass.

To do so, I'm changing a bit the internal of the PathMapping repository: instead of storing direct `path <-> filesystemPath` links in the store, it stores `path <-> filesystemPaths array`.

The problem is not present in the OptimizedPathMapping reepository as every child is resolved when added.


tests/AbstractEditableRepositoryTest.php
```diff
@@ -86,12 +86,12 @@ public function testAddFile()
...
     public function testAddMergesResourceChildren()
     {
-        $this->writeRepo->add('/webmozart/puli', $this->buildStructure($this->createDirectory('/merge', array(
+        $this->writeRepo->add('/webmozart/puli', $this->buildStructure($this->createDirectory('/foo', array(
             $this->createFile('/file1', 'original 1'),
             $this->createFile('/file2', 'original 2'),
         ))));
 
-        $this->writeRepo->add('/webmozart/puli', $this->buildStructure($this->createDirectory('/merge', array(
+        $this->writeRepo->add('/webmozart/puli', $this->buildStructure($this->createDirectory('/bar', array(
```
webmozart: I don't quite get the idea of this change. It *should* be the same path, no?

webmozart: Ah forget it, I got it now. :)

src/PathMappingRepository.php
```diff
@@ -107,7 +110,11 @@ public function listChildren($path)
      */
     public function hasChildren($path)
     {
-        return $this->getChildren($this->get($path), false)->count() !== 0;
+        if (!$this->resolveResource($path)) {
```
webmozart: IMO you should implement another method `isResourceResolvable()` which returns `true` or `false` for performance reasons. You are doing this check quite often.

webmozart: Or maybe a better solution is to extract `getFilesystemPath($path)` out of `resolveResource()` and inline `getFirstFilesystemPath()` there. Then you can use the more efficient `getFilesystemPath()` where you do boolean checks and only create resources if they are really needed. You can then remove `resolveResource()` and use this instead:

```php
$resource = $this->createResource($this->getFilesystemPath($path), $path);
```

I would rename `createAndAttachResource()` to simply `createResource()` for brevity.

tgalopin:
This will be fixed in https://github.com/puli/repository/pull/46 so I close here.

PR #41, opened 2015-08-19
Update countStore and sortStore for new key-value store architecture

tgalopin:
Quite explicit :) 

Update for https://github.com/webmozart/key-value-store/commit/2456f51dcf94562a12b8fd854f68bde8eed47c00


webmozart:
Thanks @tgalopin! :)

PR #42, opened 2015-08-20
[WIP] Add lowest and highest dependencies tests and fix errors

tgalopin:
As proposed in https://github.com/puli/issues/issues/110, it's a good idea to test the lowest and highest dependecies as well as the current ones.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,25 +1,33 @@
 language: php
 
-matrix:
-    include:
-        - php: hhvm
-        - php: 5.3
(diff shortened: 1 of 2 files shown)


tgalopin:
There i a problem with the lowest dependencies and the key-value store: https://travis-ci.org/tgalopin/repository/jobs/76506325.

I think its seomthing quite simple, just a version to update in the composer.json.

webmozart:
Awesome, thanks :)

PR #43, opened 2015-08-21
normalize path for windows machines

Big-Whoop:
Normalize path to fix windows backslash issues.

Tests are still red... This only fixes a general backslash bug, that prevents me from using puli on windows.


tgalopin:
The tests aren't passing because the lowest version of path-util needs to become 2.2. Can you update the composer.json with `"webmozart/path-util": "^2.2"`?

Big-Whoop:
This doesn't change anything because the caret operator in "^2.0" already allows V 2.2 which is already installed.

tgalopin:
Yes it does :) .

Our problem doesn't come from the high dependencies but from the low ones: https://travis-ci.org/puli/repository/builds/76626907. By settings `^2.2`, we set the low version limit, so it should work even in `deps=low` :) .

tgalopin:
Oh, I think we don't talk about the same tests ;) . The tests I'm talking about are the Travis one and you're talking about the Windows tests you're running :) . Can you update the composer.json still, so we can check the lowest dependencies?

GrahamCampbell:
This isn't need. Using the linux style slash on windows still works because php internally normalizes it before all lower level opperations.

tgalopin:
There is still a problem with the root of the filesystem (`C:/` and `/`) without the normalizing I think. Is it correct @Big-Whoop ?

webmozart:
I *think* this shouldn't be needed after webmozart/glob#5, since then RecursiveDirectoryIterator will already return normalized paths.

PR #45, opened 2015-08-24
Simplify travis.yml file

tgalopin:
As discussed with @webmozart, the travis.yml file can be simplified.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,9 +13,7 @@ matrix:
     - php: 5.5
     - php: 5.6
     - php: 5.6
-      env: deps=low
-    - php: 5.6
-      env: deps=high
+      env: COMPOSER_FLAGS='--prefer-lowest --prefer-stable'
     - php: hhvm
     - php: nightly
   allow_failures:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -23,11 +21,7 @@ matrix:
     - php: nightly
   fast_finish: true
 
-install:
-  - composer self-update -q
-  - if [ -z "$deps" ]; then composer install -n; fi;
-  - if [ "$deps" = "low" ]; then composer update --prefer-lowest --prefer-stable --prefer-source -n; fi;
-  - if [ "$deps" = "high" ]; then composer update --prefer-source -n; fi;
+install: composer update $COMPOSER_FLAGS -n
```
webmozart: This is missing the self-update, no?

GrahamCampbell: no

GrahamCampbell: You don't need to self-update?

webmozart: Why was it in there before?

GrahamCampbell: D'no. OCD?

tgalopin: According to @stof and @GrahamCampbell, self-update is not required :) 

I think it's required only if we use features of lastest versions of Composer and TravisCI version is recent enough to support `--prefer-lowest` and `--prefer-stable`. It only adds useless build time.

webmozart: :) ok

PR #49, opened 2015-09-21
Add base directory restriction on repositories

tgalopin:
Fix https://github.com/puli/issues/issues/116.

Is it what you had in mind @webmozart ?

I'm not sure about the exception to throw, I throw an InvalidArgumentException as `webmozart/assert` throw these usually.


tgalopin:
We have a problem with StyleCI: https://styleci.io/analyses/qybQoz. StyleCI thinks Resource objects are resources (as discussed in https://github.com/puli/issues/issues/119.

We should probably work on this.

tgalopin:
I think we can ignore StyleCI for the moment. If this PR if fine, I'll work on https://github.com/puli/issues/issues/117.

tgalopin:
I think this is ready. I'm starting work on #117.

src/AbstractRepository.php
```diff
@@ -39,6 +57,22 @@ protected function validateSearchLanguage($language)
...
+     * @param string $path A file or directory path.
+     *
+     * @throws InvalidArgumentException If the given path in not in the base directory.
+     */
+    protected function validatePathInBaseDirectory($path)
+    {
+        if (!Path::isBasePath($this->baseDir, $path)) {
+            throw new InvalidArgumentException(
+                sprintf('Resource "%s" is not in the repository base directory "%s"', $path, $this->baseDir)
+            );
```
webmozart: I usually format exceptions with `sprintf()`s like this to make them easier to read:

```php
throw new InvalidArgumentException(sprintf(
    'Resource "%s" is not in the repository base directory "%s"', 
    $path, 
    $this->baseDir
));
```

Could you change this?

src/FilesystemRepository.php
```diff
@@ -124,10 +119,10 @@ public static function isSymlinkSupported()
      */
     public function __construct($baseDir = '/', $symlink = true, $relative = true)
     {
-        Assert::directory($baseDir);
+        parent::__construct($baseDir);
+
         Assert::boolean($symlink);
 
-        $this->baseDir = rtrim(Path::canonicalize($baseDir), '/');
```
webmozart: Hm, this and the PathMappingRepository do two different things now. PathMappingRepository restricts newly added resources to be within the given base directory. The FilesystemRepository, on the other hand, moves added files to that directory - but the added files can be located anywhere.

Does it make sense to also restrict resources added to the FilesystemRepository to be within a specific base directory?

tgalopin: I think it make sense to have the same restriction. IMO, the problem is more should we use the same directory for restriction and base directory or should we have another option? Using the same one could be conflicting some times (for instance, the current tests wouldn't work).

What do you think?

webmozart: Using the same directory does not make sense at all. By default, for example the directory where FilesystemRepository stores its resources is `<root>/.puli/repository`, while the base directory (for security checks) is `<root>`.

I think we should remove the security checks again. Let's KISS:

* PathMappingRepository stores its paths relative to `$baseDir`
* FilesystemRepository stores its content in `$baseDir`

PR #52, opened 2015-09-23
Create LinkResource and implement it in repositories

tgalopin:
Fix https://github.com/puli/issues/issues/107.

This is a work in progress.

Small question: what happen for a LinkResource in the FilesystemRepository? How do we store it? Do we create a symlink?

### DIFF
--- a/src/AbstractPathMappingRepository.php
+++ b/src/AbstractPathMappingRepository.php
@@ -12,11 +12,13 @@
 namespace Puli\Repository;
 
 use Puli\Repository\Api\Resource\FilesystemResource;
+use Puli\Repository\Api\Resource\Resource;
 use Puli\Repository\Api\ResourceCollection;
 use Puli\Repository\Api\UnsupportedResourceException;
 use Puli\Repository\Resource\DirectoryResource;
 use Puli\Repository\Resource\FileResource;
 use Puli\Repository\Resource\GenericResource;
+use Puli\Repository\Resource\LinkResource;
 use RuntimeException;
 use Webmozart\KeyValueStore\Api\CountableStore;
 use Webmozart\KeyValueStore\Api\KeyValueStore;
@@ -53,14 +55,6 @@ abstract class AbstractPathMappingRepository extends AbstractRepository
     }
 
     /**
-     * Add the resource (internal method after checks of add()).
-     *
-     * @param string             $path
-     * @param FilesystemResource $resource
-     */
-    abstract protected function addResource($path, FilesystemResource $resource);
-
-    /**
      * {@inheritdoc}
      */
     public function add($path, $resource)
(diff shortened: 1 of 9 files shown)


tgalopin:
I implemented the LinkResource in FilesystemRepository by creating and reading symbolic links (which seems logical as it's the filesystem equivalent of a LinkResource).

What do you think of this @webmozart ?

webmozart:
Yes @tgalopin that makes sense to me :)

src/AbstractPathMappingRepository.php
```diff
@@ -79,21 +73,61 @@ public function add($path, $resource)
...
-            return;
+    /**
+     * Add the resource (internal method after checks of add()).
+     *
+     * @param string   $path
+     * @param Resource $resource
+     */
+    private function addResource($path, $resource)
+    {
+        if ($resource instanceof Resource) {
```
webmozart: You can invert the condition and throw the exception inside to reduce the code nesting level.

webmozart: i.e.

```
if (!($resource instanceof FilesystemResource || $resource instanceof LinkResource)) {
   // ...
```
```

File: src/AbstractPathMappingRepository.php [hunk 2]
```diff
@@ -173,6 +207,10 @@ protected function sortStore()
      */
     protected function createResource($filesystemPath, $path = null)
     {
+        if (0 === strpos($filesystemPath, 'k:')) {
```
webmozart: What does the `k` stand for?

src/Resource/GenericResource.php
```diff
@@ -244,4 +222,69 @@ protected function postUnserialize(array $data)
...
+    /**
+     * {@inheritdoc}
+     */
+    protected function pathGetChild($path, $relPath)
+    {
+        if (!$this->getRepository()) {
+            throw ResourceNotFoundException::forPath($path.'/'.$relPath);
+        }
+
+        return $this->getRepository()->get($path.'/'.$relPath);
```
webmozart: I don't think these additional methods are necessary. Remember that method calls are relatively expensive in PHP. IMO it doesn't make sense to reduce code duplication by moving code to an extra method if that code is relatively short and trivial (like this code here).

tests/AbstractEditableRepositoryTest.php
```diff
@@ -84,6 +85,26 @@ public function testAddFile()
         $this->assertSame(TestFile::BODY, $file->getBody());
     }
 
+    public function testAddLink()
```
webmozart: I think we should test what happens when accessing the children of a link when a link points to a directory.

webmozart:
This PR looks pretty good, thanks for working on this @tgalopin! :)

src/AbstractPathMappingRepository.php
tgalopin: I don't know, that was your proposal :D . What do you think of `link:`? It would be clearer I think.

src/Resource/GenericResource.php
tgalopin: That's a ScruntizerCI problem. It detects dupplicated code if I do copy the methods, so it fails. Another solution would be to not implement `getChild`, `hasChild`, `listChildren` and `hasChildren` in LinkResource (throw an exception for instance) and only use the target path.

webmozart: Hm ok. Let's call them `getChildOfPath()` etc. then.

tgalopin:
I made the changes. What do you think of this now?

src/AbstractPathMappingRepository.php
```diff
@@ -181,6 +213,31 @@ protected function createResource($filesystemPath, $path = null)
...
+    {
+        if ($this->contains($targetPath)) {
+            $resource = new LinkResource($targetPath);
+            $resource->attachTo($this, $path);
+
+            return $resource;
+        }
+
+        throw new RuntimeException(sprintf(
+            'Trying to create a LinkResource on a non-existing target "%s"',
```
webmozart: I don't think we should throw an exception here. We'll get an exception once the link is resolved to a target, which seem sufficient to me.

src/PathMappingRepository.php
```diff
@@ -158,23 +159,37 @@ public function hasChildren($path)
...
      * {@inheritdoc}
      */
-    protected function addResource($path, FilesystemResource $resource)
+    protected function addFilesystemResource($path, FilesystemResource $resource)
     {
-        // Don't modify resources attached to other repositories
-        if ($resource->isAttached()) {
-            $resource = clone $resource;
-        }
+        $resource->attachTo($this, $path);
```
webmozart: Could we move the call to `attachTo()` up to `addResource()`?

tests/FilesystemRepositoryCopyTest.php
```diff
@@ -178,4 +179,20 @@ public function testRemoveFailsIfLanguageNotGlob()
...
+    {
+        $this->writeRepo->add('/webmozart/link', new LinkResource('/webmozart/puli/file'));
+    }
+
+    /**
+     * @expectedException \Puli\Repository\Api\UnsupportedResourceException
+     */
+    public function testDirectoryLink()
+    {
+        $this->writeRepo->add('/webmozart/link', new LinkResource('/webmozart/puli/file'));
```
webmozart: Why does this throw an exception?

tgalopin: I use symlink to implement LinkResource in FilesystemRepository, so if symlinks are disable, LinkResource is not supported.

webmozart: Ah! Thanks :)

src/PathMappingRepository.php
tgalopin: Not easily as `OptimizedPathMappingRepository` does not attach the resource here. Do you think I should override `addResource` in `PathMappingRepository`?

src/AbstractPathMappingRepository.php
tgalopin: So `null`?

src/PathMappingRepository.php
webmozart: Why not?

src/AbstractPathMappingRepository.php
webmozart: No, simply store the LinkResource to the target path.

src/PathMappingRepository.php
tgalopin: Well, it's quite complicated just for a line of code, but I can :) .

src/AbstractPathMappingRepository.php
tgalopin: Okay.

webmozart:
Looks good to me :) Thanks!

PR #53, opened 2015-09-24
Add appveyor configuration file for CI on Windows

tgalopin:
Add CI for Windows through Appveyor.

Related to https://github.com/puli/issues/issues/115.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,31 @@
+build: false
+shallow_clone: true
+platform: x86
+clone_folder: c:\projects\puli\repository
+
+init:
+  - SET PATH=C:\Program Files\OpenSSL;c:\tools\php;%PATH%
+
+environment:
+  matrix:
+    - COMPOSER_FLAGS: ""
+    - COMPOSER_FLAGS: --prefer-lowest --prefer-stable
+
+install:
+  - cinst -y OpenSSL.Light
+  - cinst -y php
+  - cd c:\tools\php
+  - copy php.ini-production php.ini /Y
+  - echo date.timezone="UTC" >> php.ini
+  - echo extension_dir=ext >> php.ini
+  - echo extension=php_openssl.dll >> php.ini
+  - echo extension=php_mbstring.dll >> php.ini
+  - echo extension=php_fileinfo.dll >> php.ini
+  - echo memory_limit=1G >> php.ini
+  - cd c:\projects\puli\repository
+  - php -r "readfile('http://getcomposer.org/installer');" | php
+  - php composer.phar update %COMPOSER_FLAGS% --no-interaction --no-progress
+
+test_script:
+  - cd c:\projects\puli\repository
+  - vendor\bin\phpunit.bat --verbose
--- a/composer.json
+++ b/composer.json
@@ -18,7 +18,8 @@
     "require-dev": {
         "webmozart/key-value-store": "^1.0.0-beta5",
         "symfony/filesystem": "^2.0",
-        "phpunit/phpunit": "^4.6"
+        "phpunit/phpunit": "^4.6",
+        "sebastian/version": "^1.0.1"
     },
     "suggest": {
         "webmozart/key-value-store": "to use the path mapping repositories",
--- a/tests/PathMappingRepositoryTest.php
+++ b/tests/PathMappingRepositoryTest.php
@@ -224,7 +224,11 @@ class PathMappingRepositoryTest extends AbstractEditableRepositoryTest
         $this->assertEquals(str_replace(DIRECTORY_SEPARATOR, '/', __DIR__).'/Fixtures/dir5/sub', $resource->getFilesystemPath());
 
         // Find
-        $resource = $repository->find('/**/sub')->get(0);
+        $resources = $repository->find('/**/sub');
+
+        $this->assertCount(1, $resources);
+
+        $resource = $resources->get(0);
 
         $this->assertInstanceOf('Puli\Repository\Api\Resource\FilesystemResource', $resource);
         $this->assertEquals(str_replace(DIRECTORY_SEPARATOR, '/', __DIR__).'/Fixtures/dir5/sub', $resource->getFilesystemPath());
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,7 +23,7 @@ matrix:
 
 install: composer update $COMPOSER_FLAGS -n
 
-script: vendor/bin/phpunit -v --coverage-clover=coverage.clover
+script: vendor/bin/phpunit --verbose --coverage-clover=coverage.clover
 
 after_script:
   - sh -c 'if [ "$TRAVIS_PHP_VERSION" != "hhvm" ]; then wget https://scrutinizer-ci.com/ocular.phar && php ocular.phar code-coverage:upload --format=php-clover coverage.clover; fi;'


tgalopin:
Appveyor configuration seems to work well but we have a memory issue with lowest dependencies and I don't know why: https://ci.appveyor.com/project/webmozart/repository/build/job/bpqf2ox3st524vfp

Have you ever had the problem @leofeyer ? I tried to increase the memory limit but the problem stays.

leofeyer:
No, I haven't. Try to increase the memory limit even more (e.g. `1G`).

appveyor.yml
```diff
@@ -0,0 +1,41 @@
...
+  - echo extension=php_pdo_mysql.dll >> php.ini
+  - echo extension=php_soap.dll >> php.ini
+  - echo extension=php_sockets.dll >> php.ini
+  - echo extension=php_xmlrpc.dll >> php.ini
+  - echo extension=php_xsl.dll >> php.ini
+  - echo memory_limit=512M >> php.ini
+  - SET PATH=c:\tools\php;%PATH%
+  - cd c:\projects\puli\repository
+  - php -r "readfile('http://getcomposer.org/installer');" | php
+  - php composer.phar update %COMPOSER_FLAGS% --no-interaction
```
stof: you should add ``--no-progress`` because the appveyor logs don't support overwritting the output and so the composer output looks bad when it tries to display the progress

stof: @tgalopin  this comment got hidden, but it is still relevant

tgalopin:
Still have the problem :/ : https://ci.appveyor.com/project/webmozart/repository/build/job/tbko2lov39byyv6d

appveyor.yml
tgalopin: I'm doing the change :) .

tgalopin: I don't understand why https://ci.appveyor.com/project/webmozart/repository/build/1.0.4/job/8fj48ln54ld5ilo5 failed but https://ci.appveyor.com/project/webmozart/repository/build/1.0.3/job/7to83v54nqd3hw6v didn't ...

tgalopin:
I don't understand why https://ci.appveyor.com/project/webmozart/repository/build/1.0.4/job/8fj48ln54ld5ilo5 failed but https://ci.appveyor.com/project/webmozart/repository/build/1.0.3/job/7to83v54nqd3hw6v didn't. Any idea?

leofeyer:
You should maybe contact the Appveyor support.

tgalopin:
This is weird: https://ci.appveyor.com/project/webmozart/repository/build/1.0.12/job/hwy7b4ebvr78vtri

And we need to investigate on https://ci.appveyor.com/project/webmozart/repository/build/job/7igxqgvujiebbpwx (search issue, probably related to `\` vs `/`).

tgalopin:
I think we can copy `appveyor.yml` to other repositories. Can you do a last review after the final changes I did?

webmozart:
The `--prefer-lowest` build fails. Should we run `composer self-update`?

stof:
@webmozart the latest composer is downloaded for each build on AppVeyor already, as there is no pre-installed version.

stof:
found the issue: you are using ``composer install``. ``--prefer-lowest`` only exists in ``composer update``.

tgalopin:
@stof thanks, I made the change.

webmozart:
Still out of memory error.

stof:
See #55 for the testsuite failures.

webmozart:
Could you rebase this PR please?

tgalopin:
@stof @webmozart @leofeyer Do you think this `appveyor.yml` is fine and I can create PR on other repositories?

stof:
@webmozart regarding the issue on lowest deps, I think the issue is here: https://github.com/sebastianbergmann/version/blob/1.0.0/src/Version.php#L131
This code does not work fine on Windows. This was fixed in 1.0.1: https://github.com/sebastianbergmann/version/commit/da9a9cbb998f3b2cf78459dbad3037a19e759554
Given you install phpunit through composer, it means that ``--prefer-lowest`` also runs against old versions of PHPUnit and all its dependencies, which may contain bugs.

The question is whether lowest deps need to be tested on both AppVeyor and Travis. If you consider yes, you would need to force using a newer version of ``sebastianbergmann/version`` (at least 1.0.1)

webmozart:
@stof Thanks for researching this!

@tgalopin Could you add `sebastianbergmann/version` `^1.0.1` to `require-dev`?

composer.json
```diff
@@ -18,7 +18,8 @@
     "require-dev": {
         "webmozart/key-value-store": "^1.0.0-beta5",
         "symfony/filesystem": "^2.0",
-        "phpunit/phpunit": "^4.6"
+        "phpunit/phpunit": "^4.6",
+        "sebastianbergmann/version": "^1.0.1"
```
stof: sorry, the composer package name is actually ``sebastian/version`` (the namespace for composer does not match the github username)

stof:
:+1: for this PR. Fixing all tests on Windows can be considered to be a separate topic. Having a CI setup running tests on Windows will make this process easier.

appveyor.yml
```diff
@@ -0,0 +1,31 @@
...
+  - echo extension=php_mbstring.dll >> php.ini
+  - echo extension=php_fileinfo.dll >> php.ini
+  - echo memory_limit=1G >> php.ini
+  - cd c:\projects\puli\repository
+  - php -r "readfile('http://getcomposer.org/installer');" | php
+  - php composer.phar update %COMPOSER_FLAGS% --no-interaction --no-progress
+
+test_script:
+  - cd c:\projects\puli\repository
+  - vendor\bin\phpunit.bat
```
stof: should it run in verbose mode to be consistent with Travis, to output why tests are skipped ?

tests/PathMappingRepositoryTest.php
```diff
@@ -224,7 +224,11 @@ public function testResolveFilesystemResource()
         $this->assertEquals(str_replace(DIRECTORY_SEPARATOR, '/', __DIR__).'/Fixtures/dir5/sub', $resource->getFilesystemPath());
 
         // Find
-        $resource = $repository->find('/**/sub')->get(0);
+        $resources = $repository->find('/**/sub');
+
+        $this->assertEquals(1, $resources->count());
```
webmozart: `assertCount(1, $resources);`

tgalopin:
@stof Thanks a lot for your awesome work on this, you rock :) !

appveyor.yml
webmozart: sounds like a good idea

tgalopin: Yeah, I disabled it as we had way too much skipped tests and the output was not readable. As you fixed most skipped tests, we can add id back.

tests/PathMappingRepositoryTest.php
tgalopin: Good idea :D .

appveyor.yml
tgalopin: https://ci.appveyor.com/project/webmozart/repository/build/1.0.18/job/sx9weti83jmltwc5

It's better but still quite verbose. I think it's fine though.

stof: Once the testsuite is cleaned from duplicate tests, you will have only a few skipped tests for relative symlinks: https://github.com/puli/repository/pull/56#issuecomment-143433440

webmozart:
Thanks guys! :)

PR #54, opened 2015-09-25
[WIP] Implement repository versioning

tgalopin:
Fix https://github.com/puli/issues/issues/90.

This implements a ChangeStream object shared by all the repositories to log resources changes.

This PR depends on https://github.com/puli/repository/pull/52 as this introduces a new type of resource and it needs to be handled by the ChangeStream.

A few notes about this PR:

- As discussed in https://github.com/puli/issues/issues/90, I create a log stack for resources and store these logs as native PHP types (array of strings) in order to let the garbage collector clear the old resource object.
- I think the technique I'm using (creating arrays) is more performant than serializing / unserializing resources but I didn't test, what do you think?
- I created an optionnal constructor parameter for the ChangeStream in each repository. This way, developers who don't need the changes log won't enable it and the performances won't be affected at all.

This PR is for the moment a work in progress.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,7 +23,7 @@ matrix:
 
 install: composer update $COMPOSER_FLAGS -n
 
-script: vendor/bin/phpunit -v --coverage-clover=coverage.clover
+script: vendor/bin/phpunit --verbose --coverage-clover=coverage.clover
 
 after_script:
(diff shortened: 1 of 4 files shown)


webmozart:
I will review this as soon as #52 is merged.

PR #55, opened 2015-09-26
Fix tests on Windows

stof:
The repository always return paths using ``/`` as directory separator, so the expectation needs to normalize ``__DIR__`` too.

### DIFF
--- a/tests/PathMappingRepositoryTest.php
+++ b/tests/PathMappingRepositoryTest.php
@@ -221,13 +221,13 @@ class PathMappingRepositoryTest extends AbstractEditableRepositoryTest
         $resource = $repository->get('/webmozart/foo/sub');
 
         $this->assertInstanceOf('Puli\Repository\Api\Resource\FilesystemResource', $resource);
-        $this->assertEquals(__DIR__.'/Fixtures/dir5/sub', $resource->getFilesystemPath());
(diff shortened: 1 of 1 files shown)


tests/PathMappingRepositoryTest.php
```diff
@@ -221,13 +221,13 @@ public function testResolveFilesystemResource()
         $resource = $repository->get('/webmozart/foo/sub');
 
         $this->assertInstanceOf('Puli\Repository\Api\Resource\FilesystemResource', $resource);
-        $this->assertEquals(__DIR__.'/Fixtures/dir5/sub', $resource->getFilesystemPath());
+        $this->assertEquals(str_replace(__DIR__, DIRECTORY_SEPARATOR, '/').'/Fixtures/dir5/sub', $resource->getFilesystemPath());
```
tgalopin: I think you should use `Path::canonicalize` instead of `str_replace` as it's what is used in the repository.

webmozart: I think it's fine here. On a side note, for such cases (without `.`, `..` etc.) you should use `Path::normalize()` which is more efficient.

tgalopin: I need to change my appveyor PR then :D .

stof: IMO, it is better to use str_replace here in the test: it makes it clear what the expected output is.

PR #56, opened 2015-09-26
Allow running symlink-based tests on supported Windows environments

stof:
The logic to detect the symlink support on Windows is copied from the Symfony Filesystem component testsuite. The original implementation is from @nicolas-grekas.


webmozart:
Thanks! :)

stof:
@webmozart note that the filesystem repository is running the same tests many times, as you have several test classes extending ``EditableRepositoryTest`` for it, and so all tests of the parent classes are running multiple times. This may need to be improved

PR #58, opened 2015-09-26
Introduce ChangeStream and ResourceStack for resources versionning

tgalopin:
Fix https://github.com/puli/issues/issues/90.

This PR aims to introduce two concepts:
- the ChangeStream that log every resource modification on all the repositories (it's shared between repositories)
- the ResourceStack which is a ResourceCollection representing all the versions of a resource over the time

The ChangeStream, when provided in a repository, store as a simple array a representation of all the resources given to it by the repositories. Using simple arrays let PHP remove old resources objects.

If you want to retreive previous versions of a resource, you can access a classical resource and use the method `getStack()` on it:

``` php
$repository->get('/tgalopin/foo/bar')->getStack();
```

Internally, this method calls the ChangeStream to build a ResourceStack for the given resource path (here `/tgalopin/foo/bar`).

This PR is a work in progress but the main last thing to do is testing.


tgalopin:
I worked a lot on this (I restarted from scratch). Can you review this @webmozart ? It's probably not ready but as the PR is pretty big, I would like to have feedbacks as early as possible :) .

src/ChangeStream/ChangeStream.php
```diff
@@ -0,0 +1,156 @@
...
+     * @var ResourceNormalizer[]
+     */
+    private $normalizers;
+
+    /**
+     * Create a change stream.
+     */
+    public function __construct()
+    {
+        $this->stack = array();
```
EmanueleMinotto: this can be removed by replacing the attribute above with `private $stack = array();`

src/ChangeStream/ChangeStream.php
```diff
@@ -0,0 +1,156 @@
...
+     */
+    private function denormalize($data)
+    {
+        foreach ($this->normalizers as $normalizer) {
+            if (get_class($normalizer) === $data['normalizer']) {
+                return $normalizer->denormalize($data);
+            }
+        }
+
+        throw new \RuntimeException(sprintf(
```
EmanueleMinotto: shouldn't the `\RuntimeException` be included in the phpdoc?

src/ChangeStream/ChangeStream.php
tgalopin: I prefer to do the initialization in the constructor. I can change, but I think it's not that much a problem.

src/ChangeStream/ChangeStream.php
tgalopin: That's something we definitely should do, but then we should do it everywhere. For the moment, there are quite a lot of places where we didn't add the exceptions in the phpdoc.

src/AbstractEditableRepository.php
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getStack($path)
+    {
+        if (!$this->changeStream) {
+            throw new UnsupportedOperationException(sprintf(
+                'Impossible to retrieve resource stack for path "%s" as no ChangeStream is configured.',
```
webmozart: "Could not retrieve the resource stack for path "%s" as no ChangeStream was passed to the constructor of the repository."

src/AbstractEditableRepository.php
```diff
@@ -0,0 +1,80 @@
...
+            ));
+        }
+
+        return $this->changeStream->buildResourceStack($this, $path);
+    }
+
+    /**
+     * @param ChangeStream $changeStream
+     */
+    public function setChangeStream(ChangeStream $changeStream = null)
```
webmozart: Why do we allow to change the stream?

src/AbstractEditableRepository.php
```diff
@@ -0,0 +1,80 @@
...
+        $this->changeStream = $changeStream;
+    }
+
+    /**
+     * Log a change in the ChangeStream if possible.
+     *
+     * @param string       $path
+     * @param PuliResource $resource
+     */
+    public function logChange($path, PuliResource $resource)
```
webmozart: Why is this public?

src/ChangeStream/ChangeStream.php
webmozart: I agree that properties should be initialized where they are declared if possible.

src/ChangeStream/ChangeStream.php
```diff
@@ -0,0 +1,156 @@
...
+     */
+    public function __construct()
+    {
+        $this->stack = array();
+
+        $this->normalizers = array(
+            new LinkResourceNormalizer(),
+            new DirectoryResourceNormalizer(),
+            new FileResourceNormalizer(),
+            new GenericResourceNormalizer(),
```
webmozart: Why not simply serialize resources in the stream?

webmozart:
This is pretty cool! :) Thanks for working on this!

src/ChangeStream/ChangeStream.php
tgalopin: This let developers use their own resources. For instance in the `InMemoryRepository`, we support any resource implementing `PuliResource` interface so developers can use their own. This provides a way of extending the ChangeStream using normalizers. Moreover, I find it clearer and easier to maintain in the future.

src/AbstractEditableRepository.php
tgalopin: It shouldn't, I change that.

src/AbstractEditableRepository.php
tgalopin: Because I thought that maybe developers could want to either enable a new one or disable it after the repository creation. I can remove this feature if you think it should not be possible.

src/ChangeStream/ChangeStream.php
tgalopin: Ok, I'll change that :) .

src/ChangeStream/ChangeStream.php
webmozart: That exactly is why PuliResource extends Serializable :) People can adjust the serialization logic in their own implementations. If they need to inject external dependencies, we can add an initializer mechanism (similar to BindingInitializer) later on.

src/ChangeStream/ChangeStream.php
webmozart: I don't think so. I treat the documentation of exceptions similar to checked and unchecked exceptions in Java. Documented exceptions are supposed to be caught and resolved by the user of the method. Undocumented exceptions represent invalid state that cannot be resolved (read: mostly fatal errors).

We usually import global classes however, i.e. `use RuntimeException` and `throw new RuntimeException`.

src/AbstractEditableRepository.php
webmozart: I don't think this should be possible (for now).

src/ChangeStream/ChangeStream.php
tgalopin: So we simply serialize resources in the change steam ?

tgalopin: I saw your first comment, I do that :) .

src/AbstractEditableRepository.php
tgalopin: Okay.

src/ChangeStream/ChangeStream.php
webmozart: I think that's easy enough, yes. :)

src/ChangeStream/ResourceStack.php
```diff
@@ -0,0 +1,92 @@
...
+ * Represents a versionned stack of resources for a given path.
+ *
+ * You can access different versions of a resource using a ChangeStream
+ * that will return you resource stacks.
+ *
+ * @since  1.0
+ *
+ * @author Titouan Galopin <galopintitouan@gmail.com>
+ */
+class ResourceStack extends ArrayResourceCollection
```
webmozart: Is a stack really a resource collection? I doubt it.

src/ChangeStream/ChangeStream.php
```diff
@@ -0,0 +1,69 @@
...
+use Puli\Repository\Api\ResourceRepository;
+
+/**
+ * Stream to track repositories changes and fetch previous versions of resources.
+ *
+ * @since  1.0
+ *
+ * @author Titouan Galopin <galopintitouan@gmail.com>
+ */
+class ChangeStream
```
webmozart: IIUC this is not yet persistent right? IMO `ChangeStream` should be an interface in `Api`, and this one is probably an `InMemoryChangeStream`, no?

tgalopin: That's true :) .

src/ChangeStream/ResourceStack.php
tgalopin: What do you mean? The built resource stack does contains only resources objects and they are different even if they are versions of the same path.

webmozart: I mean semantically. `extends` is a "is-a" relation ship. `VW extends Car`. But a stack is not a collection, so I wouldn't extend that class here.

tgalopin: Why do you think it isn't? I personally see a resources stack as a collection, as it's a group of resources. What is a collection?

webmozart: I guess you're right.

In this case, though, do we want others to be able to manipulate the stack? (`set()`, `remove()`, ...) Or should we rather not extend anything, store the resources as array and simply provide a `toArray()` method which can be used to convert the stack to an array or a collection?

tgalopin: You're right, I didn't think very much about what we should expose here. I think a resource stack should be immutable and use a toArray as you suggested. I'll made the change :) .

tgalopin:
I updated this.

src/Api/ChangeStream/ResourceStack.php
```diff
@@ -0,0 +1,64 @@
...
+ * Represents a versionned stack of resources for a given path.
+ *
+ * You can access different versions of a resource using a ChangeStream
+ * that will return you resource stacks.
+ *
+ * @since  1.0
+ *
+ * @author Titouan Galopin <galopintitouan@gmail.com>
+ */
+interface ResourceStack
```
webmozart: Given that the `ResourceStack` is a basic domain object that is not likely to be replaced, I think it would be sufficient to simply add the implementation here (just like with `ResourceMetadata`). This also enables us to add more methods, should we need any.

tgalopin: Okay :) .

src/Resource/GenericResource.php
```diff
@@ -125,6 +125,18 @@ public function listChildren()
     /**
      * {@inheritdoc}
      */
+    public function getStack()
+    {
+        if (!$this->getRepository()) {
+            throw ResourceNotFoundException::forPath($this->getRepositoryPath());
```
webmozart: Is this the right exception?

tests/AbstractEditableRepositoryTest.php
```diff
@@ -488,4 +492,31 @@ public function testLinkTargetWithoutRepository()
...
+        $this->writeRepo = $this->createWriteRepository($stream);
+        $this->writeRepo->add('/path', $v1 = new FileResource(__DIR__.'/Fixtures/dir1/file1'));
+        $this->writeRepo->add('/path', $v2 = new FileResource(__DIR__.'/Fixtures/dir1/file2'));
+        $this->writeRepo->add('/path', $v3 = new FileResource(__DIR__.'/Fixtures/dir2/file2'));
+
+        $resourceStack = $this->writeRepo->getStack('/path');
+
+        $this->assertInstanceOf('Puli\Repository\Api\ChangeStream\ResourceStack', $resourceStack);
+        $this->assertCount(3, $resourceStack->getAvailableVersions());
+        $this->assertFileResourcesEquals($v1, $resourceStack->getFirstVersion());
```
webmozart: Why can't you use `assertEquals()` here?

tests/Resource/TestNullResource.php
```diff
@@ -0,0 +1,140 @@
...
+
+namespace Puli\Repository\Tests\Resource;
+
+use Puli\Repository\Api\Resource\PuliResource;
+use Puli\Repository\Api\ResourceRepository;
+
+/**
+ * @author Titouan Galopin <galopintitouan@gmail.com>
+ */
+class TestNullResource implements PuliResource
```
webmozart: Is this class used?

webmozart:
This looks really good! :)

I would change the verb "log" to "append" for the stream. A stream is something continuous and we are appending stuff at the end, so "append" seems easier to understand to me.

Did you test all the exceptions you are throwing?

src/AbstractEditableRepository.php
```diff
@@ -0,0 +1,73 @@
...
+        return $this->changeStream->buildResourceStack($this, $path);
+    }
+
+    /**
+     * Log a change in the ChangeStream if possible.
+     *
+     * @param string       $path
+     * @param PuliResource $resource
+     */
+    protected function logChange($path, PuliResource $resource)
```
webmozart: You can rename this to `appendToChangeStream()` which makes its use clearer in the sub-classes.

webmozart: Also, since the `$path` can be accessed from the `$resource`, can't we remove that argument?

src/ChangeStream/InMemoryChangeStream.php
```diff
@@ -0,0 +1,62 @@
...
+    /**
+     * {@inheritdoc}
+     */
+    public function log($path, PuliResource $resource)
+    {
+        if (!array_key_exists($path, $this->stack)) {
+            $this->stack[$path] = array();
+        }
+
+        $this->stack[$path][] = serialize($resource);
```
webmozart: I don't think you need to serialize/unserialize resources for the InMemoryChangeStream. I don't see any benefit here.

src/ChangeStream/InMemoryChangeStream.php
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function buildResourceStack(ResourceRepository $repository, $path)
+    {
+        $stack = array();
+
+        if (isset($this->stack[$path]) && is_array($this->stack[$path])) {
```
webmozart: Is the `is_array()` check necessary? How can this key be set but not an array?

src/Resource/GenericResource.php
tgalopin: I don't know ... It's just the same that is thrown on other methods like this one.

webmozart: Not really. I'd rather return a stack here that contains just this resource.

tests/AbstractEditableRepositoryTest.php
tgalopin: It's just that the created resource is not associated to the repository so it doesn't have a path. However, on call on `getStack`, the resources found are associated to the current repo and path. So the objects and not exactly the same.

tests/Resource/TestNullResource.php
tgalopin: No, I'll remove it.

src/Resource/GenericResource.php
tgalopin: Okay, let's do this.

src/ChangeStream/InMemoryChangeStream.php
tgalopin: That's totally true. I'll change that.

src/AbstractEditableRepository.php
tgalopin: We need the $path as the FilesystemRepository does not attach resources to itself on adding.

src/AbstractEditableRepository.php
```diff
@@ -0,0 +1,73 @@
...
+ *
+ * @author Bernhard Schussek <bschussek@gmail.com>
+ * @author Titouan Galopin <galopintitouan@gmail.com>
+ */
+abstract class AbstractEditableRepository extends AbstractRepository implements EditableRepository
+{
+    /**
+     * @var ChangeStream
+     */
+    protected $changeStream;
```
webmozart: This can be private no?

src/ChangeStream/ResourceStack.php
```diff
@@ -0,0 +1,113 @@
...
+ */
+
+namespace Puli\Repository\ChangeStream;
+
+use Puli\Repository\Api\Resource\PuliResource;
+use RuntimeException;
+use Webmozart\Assert\Assert;
+
+/**
+ * Represents a versionned stack of resources for a given path.
```
webmozart: "versioned"

src/ChangeStream/ResourceStack.php
```diff
@@ -0,0 +1,113 @@
...
+
+        $this->stack = $stack;
+    }
+
+    /**
+     * Get the last version from the stack.
+     *
+     * @return PuliResource
+     */
+    public function getCurrentVersion()
```
webmozart: Right now, the stack's wording is a bit ambiguous. `getCurrentVersion()` returns a resource, `getVersion()` too but `getAvailableVersions()` returns integers.

I think this class would be more useful if we do:

* `getCurrentVersion()` returns the latest version number
* `getVersions()` returns all version numbers
* `getCurrent()` returns the resource for the current version
* `getFirst()` returns the resource for the first version
* `get($version)` returns the resource for a specific version
* `toArray()` returns all resources indexed by version

Then "version" consistently refers to the integer number that corresponds to a resource at a specific point in time.

tgalopin:
I improved a few tests and fixed various coding style and other namings. I also squashed the commits. What do you think of this @webmozart?

src/AbstractEditableRepository.php
```diff
@@ -0,0 +1,72 @@
...
+    {
+        $this->changeStream = $changeStream;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getStack($path)
+    {
+        if (!$this->changeStream) {
```
webmozart: Shouldn't we rather rely on the parent implementation here, i.e. return a stack of size 1?

src/ChangeStream/ResourceStack.php
```diff
@@ -0,0 +1,139 @@
...
+{
+    /**
+     * @var array
+     */
+    private $stack;
+
+    /**
+     * @param array $stack
+     */
+    public function __construct($stack)
```
webmozart: I think you should add an `array` type hint here rather than the assertion below.

src/ChangeStream/ResourceStack.php
```diff
@@ -0,0 +1,139 @@
...
+    {
+        Assert::isArray($stack, 'Built resource stack must be an array.');
```
webmozart: You can rather add the assertion `allIsInstanceOf('Puli\Repository\Api\PuliResource')` here.

src/ChangeStream/ResourceStack.php
```diff
@@ -0,0 +1,139 @@
...
+
+        $this->stack = $stack;
```
webmozart: You should do `$this->stack = array_values($stack);` here to ensure that the arrays is 0-indexed.

src/ChangeStream/ResourceStack.php
webmozart: Also you can add `Assert::greaterThanEq(1, count($stack), ...)` to ensure that we have at least one resource.

src/ChangeStream/ResourceStack.php
```diff
@@ -0,0 +1,139 @@
...
+    }
+
+    /**
+     * Get the current version number.
+     *
+     * @return int
+     */
+    public function getCurrentVersion()
+    {
+        $versions = $this->getVersions();
```
webmozart: There's no need to copy the `$stack` array around in memory as much. Note that most of the `array_*()` functions create copies of the array that is passed to them.

I'd rather do this:

```php
return count($this->stack);
```

There's no need for the count check here if we do it in the constructor.

src/ChangeStream/ResourceStack.php
```diff
@@ -0,0 +1,139 @@
...
+    }
+
+    /**
+     * Get the first version number.
+     *
+     * @return int
+     */
+    public function getFirstVersion()
+    {
+        $versions = $this->getVersions();
```
webmozart: You can simply `return 1;` here.

src/ChangeStream/ResourceStack.php
```diff
@@ -0,0 +1,139 @@
...
+     *
+     * @param int $version The version number (first is 0).
+     *
+     * @return PuliResource
+     */
+    public function get($version)
+    {
+        $versions = $this->getVersions();
+
+        if (0 === count($versions)) {
```
webmozart: This check can be removed. Instead, check:

```php
if (!isset($this->stack[$version])) {
    throw new OutOfBoundsException(...);
}
```

src/ChangeStream/ResourceStack.php
webmozart: These assertions should be tested, of course. :)

src/ChangeStream/ResourceStack.php
tgalopin: That's something interesting: for you, the first version is 1 and the last is count($stack) ? For the moment I started at 0. What do you think is the most intuitive for developers?

webmozart: I can't really say. For me, versions intuitively start at 1, but I can't give a strong reason why.

src/ChangeStream/ResourceStack.php
tgalopin: It's `Assert::greaterThanEq(count($stack), 1, 'A ResourceStack cannot be empty');` (wasn't easy to debug :D ).

webmozart: Oops, yes, sorry :)

src/ChangeStream/ResourceStack.php
tgalopin: I'll keep it like this for the moment as it follows array conventions across languages, but it's open to debate :) .

tgalopin:
I updated this.

src/ChangeStream/InMemoryChangeStream.php
```diff
@@ -0,0 +1,62 @@
...
+
+            foreach ($this->stack[$path] as $resource) {
+                $resource->attachTo($repository, $path);
+                $resources[] = $resource;
+            }
+
+            return new ResourceStack($resources);
+        }
+
+        throw new \InvalidArgumentException(sprintf('No version of path %s were found in the ChangeStream', $path));
```
webmozart: Can you remove the slash and import the class?

src/ChangeStream/InMemoryChangeStream.php
```diff
@@ -0,0 +1,62 @@
...
+     * @var array
+     */
+    private $stack = array();
+
+    /**
+     * {@inheritdoc}
+     */
+    public function append(PuliResource $resource)
+    {
+        if (!array_key_exists($resource->getPath(), $this->stack)) {
```
webmozart: You can use `isset` here like in buildStack()

src/ChangeStream/InMemoryChangeStream.php
```diff
@@ -0,0 +1,62 @@
...
+
+        $this->stack[$resource->getPath()][] = $resource;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function buildStack(ResourceRepository $repository, $path)
+    {
+        if (isset($this->stack[$path])) {
```
webmozart: I'd invert the condition and move the exception in here. Short if statements generally increase readability.

tgalopin:
I made the changes.

src/ChangeStream/ResourceStack.php
```diff
@@ -0,0 +1,121 @@
...
+    }
+
+    /**
+     * Get the first version number.
+     *
+     * @return int
+     */
+    public function getFirstVersion()
+    {
+        return 0;
```
webmozart: Thought about this a bit more but I still think it's better to have the first version 1. For collection indices, starting with 0 makes sense to me, but for versions I think it's weird. Alone the fact that "get first version" (1st) returns 0 is an indicator for me. :)

src/ChangeStream/ResourceStack.php
```diff
@@ -0,0 +1,121 @@
...
+
+    /**
+     * @param array $stack
+     */
+    public function __construct(array $stack)
+    {
+        Assert::allIsInstanceOf($stack, 'Puli\Repository\Api\Resource\PuliResource');
+        Assert::greaterThanEq(count($stack), 1, 'A ResourceStack cannot be empty');
+
+        $this->stack = array_values($stack);
```
webmozart: You can change this to `array_combine(range(1, count($stack)), $stack)` in order to reindex the stack starting at 1.

src/ChangeStream/ResourceStack.php
webmozart: On the other hand, whenever we deal with arrays, we start with zero, so starting with 1 might complicate processing a list of versions outside of this class. Let's leave this as is.

webmozart:
Awesome sauce :tada: Merging! :+1:

harikt:
2015 assets ;) . Thank you.

PR #59, opened 2015-09-26
Add Composer directory to Appveyor cache

tgalopin:
Because tests are better when they are fast :) .


tgalopin:
As explained by @stof here: https://github.com/puli/manager/pull/9#discussion-diff-40499480, I restricted the Composer cache to the `files` directory, even for Travis.

stof:
:+1:

PR #61, opened 2015-09-28
[WIP] Store relative paths in path mapping repositories

tgalopin:
Fix https://github.com/puli/issues/issues/117.

This PR aims to implement a base directory for `PathMappingRepository` and `OptimizedPathMappingRepository` in order to store relative paths in the underlying key-value stores.

Here is the plan:
1. Put all the fixtures (even the temporary ones) in the `tests/Fixtures` directory **done**
2. Add testing for base directory check and stores **done**
3. Implement the feature to fix the tests **done**

### DIFF
--- a/src/AbstractPathMappingRepository.php
+++ b/src/AbstractPathMappingRepository.php
@@ -43,13 +43,20 @@ abstract class AbstractPathMappingRepository extends AbstractRepository
     protected $store;
 
     /**
+     * @var string
+     */
+    protected $baseDirectory;
+
+    /**
      * Creates a new repository.
      *
-     * @param KeyValueStore $store The store of all the paths.
+     * @param KeyValueStore $store
+     * @param string        $baseDirectory The store of all the paths.
      */
-    public function __construct(KeyValueStore $store)
+    public function __construct(KeyValueStore $store, $baseDirectory)
     {
         $this->store = $store;
+        $this->baseDirectory = $baseDirectory;
 
         $this->createRoot();
     }
@@ -101,6 +108,16 @@ abstract class AbstractPathMappingRepository extends AbstractRepository
         if ($resource instanceof LinkResource) {
             $this->addLinkResource($path, $resource);
         } else {
+            // Check resource is in base directory
+            if (!Path::isBasePath($this->baseDirectory, $resource->getFilesystemPath())) {
+                throw new UnsupportedResourceException(sprintf(
+                    'This path-mapping repository is based on directory "%s" and '.
+                    'therefore does not support filesystem resource "%s"',
+                    $this->baseDirectory,
+                    $resource->getFilesystemPath()
+                ));
+            }
+
             $this->addFilesystemResource($path, $resource);
         }
     }
@@ -201,12 +218,18 @@ abstract class AbstractPathMappingRepository extends AbstractRepository
      */
     protected function createResource($filesystemPath, $path = null)
     {
+        // Link resource
         if (0 === strpos($filesystemPath, 'l:')) {
             return $this->createLinkResource(substr($filesystemPath, 2), $path);
         }
 
-        if ($filesystemPath && file_exists($filesystemPath)) {
-            return $this->createFilesystemResource($filesystemPath, $path);
+        // Filesystem resource
+        if (is_string($filesystemPath)) {
+            $filesystemPath = $this->makePathAbsolute($filesystemPath);
+
+            if (file_exists($filesystemPath)) {
+                return $this->createFilesystemResource($filesystemPath, $path);
+            }
         }
 
         return $this->createVirtualResource($path);
@@ -277,5 +300,50 @@ abstract class AbstractPathMappingRepository extends AbstractRepository
         $resource->attachTo($this, $path);
 
         return $resource;
+    }
+
+    /**
+     * Transform a relative path into an absolute path.
+     *
+     * @param string $absolutePath
+     *
+     * @return string
+     */
+    protected function makePathRelative($absolutePath)
+    {
+        return Path::makeRelative($absolutePath, $this->baseDirectory);
+    }
+
+    /**
+     * Transform a relative path into an absolute path.
+     *
+     * @param string $relativePath
+     *
+     * @return string
+     */
+    protected function makePathAbsolute($relativePath)
+    {
+        if (0 === strpos($relativePath, 'l:')) {
+            // Link
+            return $relativePath;
+        }
+
+        return Path::makeAbsolute($relativePath, $this->baseDirectory);
+    }
+
+    /**
+     * Transform a collection of relative paths into a collection of absolute paths.
+     *
+     * @param array $relativePaths
+     *
+     * @return array
+     */
+    protected function makePathsAbsolute($relativePaths)
+    {
+        foreach ($relativePaths as $key => $relativePath) {
+            $relativePaths[$key] = $this->makePathAbsolute($relativePath);
+        }
+
+        return $relativePaths;
     }
 }
--- a/src/AbstractPathMappingRepository.php
+++ b/src/AbstractPathMappingRepository.php
@@ -50,7 +50,7 @@ abstract class AbstractPathMappingRepository extends AbstractRepository
     /**
      * Creates a new repository.
      *
-     * @param KeyValueStore $store
+     * @param KeyValueStore $store         The store of all the paths.
      * @param string        $baseDirectory The store of all the paths.
      */
     public function __construct(KeyValueStore $store, $baseDirectory)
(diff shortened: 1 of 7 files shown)


tgalopin:
This PR seems to be working properly. Can you review @webmozart?

tgalopin:
Well, I spoke too fast, it does not work on Windows. I'll work on that.

src/AbstractPathMappingRepository.php
```diff
@@ -278,4 +299,54 @@ protected function createVirtualResource($path = null)
...
+    /**
+     * Transform a collection of relative paths into a collection of absolute paths.
+     *
+     * @param array $relativePaths
+     *
+     * @return array
+     */
+    protected function makePathsCollectionAbsolute($relativePaths)
+    {
+        foreach ($relativePaths as &$relativePath) {
```
webmozart: I'd rather include the index in the loop and then assign by index rather than by reference. Using references in loops can cause bugs that are hard to spot.

src/AbstractPathMappingRepository.php
```diff
@@ -278,4 +299,54 @@ protected function createVirtualResource($path = null)
...
+    }
+
+    /**
+     * Transform a collection of relative paths into a collection of absolute paths.
+     *
+     * @param array $relativePaths
+     *
+     * @return array
+     */
+    protected function makePathsCollectionAbsolute($relativePaths)
```
webmozart: `makePathsAbsolute()` should suffice

src/AbstractPathMappingRepository.php
```diff
@@ -278,4 +299,54 @@ protected function createVirtualResource($path = null)
...
+    /**
+     * Transform a relative path into an absolute path.
+     *
+     * @param string $absolutePath
+     *
+     * @return string
+     */
+    protected function makePathRelative($absolutePath)
+    {
+        if (0 === strpos($absolutePath, 'l:')) {
```
webmozart: Why can this be called with a link?

src/AbstractPathMappingRepository.php
```diff
@@ -205,8 +222,12 @@ protected function createResource($filesystemPath, $path = null)
             return $this->createLinkResource(substr($filesystemPath, 2), $path);
         }
 
-        if ($filesystemPath && file_exists($filesystemPath)) {
-            return $this->createFilesystemResource($filesystemPath, $path);
+        if ($filesystemPath) {
```
webmozart: Can you make this check more concrete?

src/AbstractPathMappingRepository.php
```diff
@@ -101,6 +108,16 @@ private function addResource($path, $resource)
         if ($resource instanceof LinkResource) {
             $this->addLinkResource($path, $resource);
         } else {
+            // Check resource is in base directory
+            if (!Path::isBasePath($this->baseDirectory, $resource->getFilesystemPath())) {
+                throw new UnsupportedResourceException(sprintf(
+                    'This PathMapping repository is based on directory "%s" and '.
```
webmozart: This path-mapping repository...

src/AbstractPathMappingRepository.php
tgalopin: Because under Windows, `l:/directory` will be understood as a partition where it shouldn't. It avoids problems.

webmozart: I mean, why is this method called at all when adding links?

src/AbstractPathMappingRepository.php
tgalopin: Good idea :) .

src/AbstractPathMappingRepository.php
tgalopin: It isn't :D . I change that.

src/AbstractPathMappingRepository.php
```diff
@@ -43,13 +43,20 @@
...
     /**
+     * @var string
+     */
+    protected $baseDirectory;
+
+    /**
      * Creates a new repository.
      *
-     * @param KeyValueStore $store The store of all the paths.
+     * @param KeyValueStore $store
```
webmozart: You removed the description here by accident.

src/AbstractPathMappingRepository.php
```diff
@@ -101,6 +108,16 @@ private function addResource($path, $resource)
         if ($resource instanceof LinkResource) {
             $this->addLinkResource($path, $resource);
         } else {
+            // Check resource is in base directory
+            if (!Path::isBasePath($this->baseDirectory, $resource->getFilesystemPath())) {
+                throw new UnsupportedResourceException(sprintf(
+                    'This path-mapping repository is based on directory "%s" and '.
+                    'therefore does not support filesystem resource "%s"',
```
webmozart: I would reword this exception message to be more easily understandable:

> Can only add resources in %s. Tried to add %s.

src/AbstractPathMappingRepository.php
```diff
@@ -278,4 +301,49 @@ protected function createVirtualResource($path = null)
...
+    /**
+     * Transform a relative path into an absolute path.
+     *
+     * @param string $relativePath
+     *
+     * @return string
+     */
+    protected function makePathAbsolute($relativePath)
+    {
+        if (0 === strpos($relativePath, 'l:')) {
```
webmozart: Is this method ever called with a link?

webmozart: Ah I see it can. I would rename this method (and the plural version before) to `resolveRelativePath()`.

tests/OptimizedPathMappingRepositoryTest.php
```diff
@@ -170,4 +105,33 @@ public function testRemoveFailsIfLanguageNotGlob()
...
+    {
+        $repo = $this->createBaseDirectoryRepository($this->store, __DIR__.'/Fixtures');
+        $repo->add('/webmozart/file', new FileResource(__DIR__.'/Fixtures/dir1/file1'));
+        $repo->add('/webmozart/dir', new DirectoryResource(__DIR__.'/Fixtures/dir2'));
+
+        $this->assertTrue($repo->contains('/webmozart/file'));
+        $this->assertTrue($repo->contains('/webmozart/dir'));
+        $this->assertEquals('dir1/file1', $this->store->get('/webmozart/file'));
+        $this->assertEquals('dir2', $this->store->get('/webmozart/dir'));
+        $this->assertPathsEquals(__DIR__.'/Fixtures/dir1/file1', $repo->get('/webmozart/file')->getFilesystemPath());
```
webmozart: `getFilesystemPath()` should already return a normalized path. Hence it should be sufficient to do:

```php
$this->assertSame(Path::normalize(__DIR__.'/Fixtures/dir1/file1'), $repo->get('/webmozart/file')->getFilesystemPath());
```

This is also the way I solved failing Windows tests in other repositories.

You can remove `assertPathsEquals()`.

webmozart: It's probably even easier to add a property `$fixturesDir` to the test and set it in `setUp()`:

```php
$this->fixturesDir = Path::normalize(__DIR__.'/Fixtures');
```

Then you don't need to call `normalize()` in the individual tests.

tests/PathMappingRepositoryTest.php
```diff
@@ -17,71 +17,30 @@
...
-
-        $filesystem = new Filesystem();
-        $filesystem->remove($this->tempDir);
+        return new PathMappingRepository($store, $baseDirectory);
     }
 
     protected function createPrefilledRepository(Resource $root)
     {
-        $repo = new PathMappingRepository(new ArrayStore());
+        $repo = new PathMappingRepository(new ArrayStore(), ('\\' === DIRECTORY_SEPARATOR) ? 'C:/' : '/');
```
webmozart: Can we guarantee that this is always `C:/`? This smells like a volatile test to me.

tgalopin:
All green \o/ (too rare to not notice :D ).

src/AbstractPathMappingRepository.php
```diff
@@ -101,6 +108,15 @@ private function addResource($path, $resource)
         if ($resource instanceof LinkResource) {
             $this->addLinkResource($path, $resource);
         } else {
+            // Check resource is in base directory
+            if (!Path::isBasePath($this->baseDirectory, $resource->getFilesystemPath())) {
```
webmozart: You can coalesce this into an `if - elseif - else`

src/AbstractPathMappingRepository.php
```diff
@@ -278,4 +300,49 @@ protected function createVirtualResource($path = null)
...
+    /**
+     * Transform a relative path into an absolute path.
+     *
+     * @param string $absolutePath
+     *
+     * @return string
+     */
+    protected function makePathRelative($absolutePath)
+    {
+        return Path::makeRelative($absolutePath, $this->baseDirectory);
```
webmozart: Let's remove this method and use `Path::makeRelative()` directly. Less method calls = better (in PHP).

src/AbstractPathMappingRepository.php
```diff
@@ -278,4 +300,49 @@ protected function createVirtualResource($path = null)
...
+            return $relativePath;
+        }
+
+        return Path::makeAbsolute($relativePath, $this->baseDirectory);
+    }
+
+    /**
+     * Transform a collection of relative paths into a collection of absolute paths.
+     *
+     * @param array $relativePaths
```
webmozart: `string[]`

src/AbstractPathMappingRepository.php
```diff
@@ -278,4 +300,49 @@ protected function createVirtualResource($path = null)
...
+     *
+     * @return array
```
webmozart: `string[]`

PR #64, opened 2015-10-05
Implement auto-resolvment of directories links

tgalopin:
Fix https://github.com/puli/issues/issues/125.

As suggested by @webmozart, it should be possible to access the children of a directory link seemlessly, ie something like:

``` php
$repo = new InMemoryRepository();
$repo->add('/dir', new DirectoryResource('/some/directory'));
$repo->add('/dir-link', new LinkResource('/dir'));
$repo->get('/dir-link/file');
```

This PR aims to implement this.


tests/AbstractEditableRepositoryTest.php
```diff
@@ -473,4 +473,20 @@ public function testDirectoryLink()
         $this->assertSame($this->readRepo, $target->getRepository());
         $this->assertCount(2, $target->listChildren());
     }
+
+    public function testDirectoryAutoresolveLinkChildren()
+    {
```
webmozart: Can you move this test to `AbstractRepositoryTest`?

src/InMemoryRepository.php
```diff
@@ -261,6 +261,26 @@ private function removeResource(Resource $resource)
     }
 
     /**
+     * {@inheritdoc}
+     */
+    protected function resolveWithLinks($path)
+    {
+        foreach ($this->resources as $resource) {
```
webmozart: I'd maintain a separate list `$links` that keeps a reference of all `LinkResources`. `$resources` can be quite large.

src/InMemoryRepository.php
```diff
@@ -261,6 +261,26 @@ private function removeResource(Resource $resource)
     }
 
     /**
+     * {@inheritdoc}
+     */
+    protected function resolveWithLinks($path)
```
webmozart: `resolveLinks()`?

src/InMemoryRepository.php
```diff
@@ -261,6 +261,26 @@ private function removeResource(Resource $resource)
...
+    {
+        foreach ($this->resources as $resource) {
+            $resourcePath = rtrim($resource->getPath(), '/');
+
+            if ($resource instanceof LinkResource && 0 === strpos($path, $resourcePath)) {
```
webmozart:
```php
0 === strpos($path, $linkPath.'/')
```

You don't want to catch false positives.

src/InMemoryRepository.php
```diff
@@ -261,6 +261,26 @@ private function removeResource(Resource $resource)
     }
 
     /**
+     * {@inheritdoc}
+     */
+    protected function resolveWithLinks($path)
+    {
+        foreach ($this->resources as $resource) {
+            $resourcePath = rtrim($resource->getPath(), '/');
```
webmozart: `$linkPath`

src/InMemoryRepository.php
```diff
@@ -261,6 +261,26 @@ private function removeResource(Resource $resource)
...
+
+            if ($resource instanceof LinkResource && 0 === strpos($path, $resourcePath)) {
+                $realpath = substr_replace($path, rtrim($resource->getTargetPath(), '/'), 0, strlen($resourcePath));
+
+                if (array_key_exists($realpath, $this->resources)) {
```
webmozart: What if the target is a link again? Or in a linked directory?

webmozart:
Added some comments above. What happens when adding resources to a linked directory? I think these should be added to the linked path as well.

You need to rebase this on master btw. as `Resource` has changed to `PuliResource` now.

tgalopin:
I fixed a lot of issues and reorganized the code. Can you review this @webmozart?

EmanueleMinotto:
What about a relative links check in the `AbstractPathMappingRepository`?
I see a lot of `strpos(..., 'l:')` and an abstraction could be an idea considering the convention.

tgalopin:
That's a good idea. A few static helpers methods in the AbstractPathMappingRepository could be useful, like `AbstractPathMappingRepository::isPathMappingLink` and `AbstractPathMappingRepository::getPathMappingLinkTarget`.What do you think?

EmanueleMinotto:
@tgalopin seems good :+1:

tgalopin:
I did this :) .

src/AbstractRepository.php
```diff
@@ -52,4 +53,29 @@ protected function sanitizePath($path)
 
         return Path::canonicalize($path);
     }
+
+    /**
+     * @param string $path
+     *
+     * @return resource
```
webmozart: This return type seems to be wrong

src/AbstractRepository.php
```diff
@@ -52,4 +53,29 @@ protected function sanitizePath($path)
...
+
+        throw ResourceNotFoundException::forPath($path);
+    }
+
+    /**
+     * Resolve links in the given path to find the resource.
+     *
+     * @param string $path
+     *
+     * @return resource|null Return the found resource or null if nothing was found.
```
webmozart: same here

src/InMemoryRepository.php
```diff
@@ -256,11 +269,55 @@ private function removeResource(PuliResource $resource)
 
         unset($this->resources[$path]);
 
+        if ($resource instanceof LinkResource && array_key_exists($path, $this->links)) {
```
webmozart: You can change this to `isset($this->links[$path])` which performs slightly better.

src/InMemoryRepository.php
```diff
@@ -256,11 +269,55 @@ private function removeResource(PuliResource $resource)
...
+
+    /**
+     * Find the real target of a given link (resolve all the links until
+     * a non-link resource is found).
+     *
+     * @param LinkResource $linkResource
+     *
+     * @return PuliResource
+     */
+    private function getRealTarget(LinkResource $linkResource)
```
webmozart: You could add this logic to `LinkResource::getTarget()` and activate it through a `$followLinks` argument (`false` by default).

src/AbstractPathMappingRepository.php
```diff
@@ -330,4 +330,48 @@ protected function resolveRelativePaths($relativePaths)
...
     }
+
+    /**
+     * Check if the given target path represents a link.
+     *
+     * @param string $targetPath
+     *
+     * @return bool
+     */
+    public static function isLinkTarget($targetPath)
```
webmozart: Should these methods really be public?

tgalopin: Not really, and not even static. I initially used them in the tests but not anymore. I'll set them as private.

tests/AbstractRepositoryTest.php
```diff
@@ -718,4 +765,78 @@ public function testFindExpectsStringPath()
 
         $repo->find(new \stdClass());
     }
+
+    public function testAutoresolveLinks()
```
webmozart: You should split this test into separate tests and give them descriptive names, then you can spare the comments inside. As it is right now, a failing test does not say anything about the actual problem ("autoresolve links fails").

src/InMemoryRepository.php
tgalopin: Good idea, I will do that.

src/InMemoryRepository.php
tgalopin: What's the difference?

tests/AbstractRepositoryTest.php
webmozart: Example: `testGetChildOfDirectoryLink()`

src/InMemoryRepository.php
webmozart: `array_key_exists()` is a function, `isset()` is a language construct. They behave slightly differently (`isset()` returns `false` for `null` values while `array_key_exists()` returns `true`) but that doesn't matter here.

webmozart:
Thanks for working on this! :) Please remember to fix the tests on Windows when you're done.

src/InMemoryRepository.php
tgalopin: Okay, thanks for the info :) .

src/AbstractRepository.php
```diff
@@ -52,4 +54,29 @@ protected function sanitizePath($path)
 
         return Path::canonicalize($path);
     }
+
+    /**
```
webmozart: I think you should add some PhpDoc with examples here that describes what exactly this method does.

webmozart: Maybe also for the method below. These can be tricky to understand.

src/AbstractRepository.php
```diff
@@ -52,4 +54,29 @@ protected function sanitizePath($path)
...
+     * @param string $path
+     *
+     * @return PuliResource
+     */
+    protected function followLinksOrFail($path)
+    {
+        $resolvedWithLinks = $this->followLinks($path);
```
webmozart: If I understand correctly, this method only follows one level of links, no? I.e., if a link points to another link (hypothetically), this method will return the second link, not the final target?

tests/AbstractRepositoryTest.php
```diff
@@ -718,4 +774,100 @@ public function testFindExpectsStringPath()
...
+    public function testLinkDirectory()
+    {
+        $repo = $this->createAutoresolveLinksRepository();
+
+        // Real directory
+        /** @var DirectoryResource $realdir */
+        $realdir = $repo->get('/webmozart/realdir');
+
+        $this->assertInstanceOf('Puli\Repository\Resource\DirectoryResource', $realdir);
+        $this->assertPathsAreEqual(__DIR__.'/Fixtures/dir1', $realdir->getFilesystemPath());
```
webmozart: These test methods should be split up in separate test methods. A test method generally consists of several steps:

* fixture setup
* execution of the test
* assertions

The fact that you repeat these steps here shows that this test is testing too much (test smell). The result in practice is that the test doesn't document well what functionality it is testing.

tests/AbstractRepositoryTest.php
```diff
@@ -718,4 +774,100 @@ public function testFindExpectsStringPath()
 
         $repo->find(new \stdClass());
     }
+
+    public function testLinkDirectory()
+    {
+        $repo = $this->createAutoresolveLinksRepository();
```
webmozart: I'm not a fan of this creation method. The method largely hides what the test is actually testing. If I read the test, I wonder why the assertions below should hold, and the reason lies in the fixture setup (which link points to where, ...). I think you should inline this method everywhere and reduce it to the minimum required for each test.

src/AbstractRepository.php
tgalopin: The `followLinks` method fully resolves the links so it will return the final target. I'll add this in the phpdoc.

tests/AbstractRepositoryTest.php
tgalopin: Okay, I'll try to find a better way to tests this stuff :) .

webmozart:
Closing this as I am refactoring the path mapping repositories and fixing these problems along the way.

PR #65, opened 2015-10-05
Create helper to return directly target of link resources

tgalopin:
Fix https://github.com/puli/issues/issues/126.

### DIFF
--- a/src/Resource/LinkResource.php
+++ b/src/Resource/LinkResource.php
@@ -52,6 +52,18 @@ class LinkResource extends GenericResource implements Resource
     /**
      * {@inheritdoc}
      */
+    public function getTarget()
+    {
+        if (!$this->getRepository()) {
+            throw ResourceNotFoundException::forPath($this->getTargetPath());
+        }
+
+        return $this->getRepository()->get($this->getTargetPath());
+    }
+
+    /**
+     * {@inheritdoc}
+     */
     public function getChild($relPath)
     {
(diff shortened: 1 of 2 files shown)


src/Resource/LinkResource.php
```diff
@@ -52,6 +52,18 @@ public function getTargetPath()
     /**
      * {@inheritdoc}
      */
+    public function getTarget()
+    {
+        if (!$this->getRepository()) {
+            throw ResourceNotFoundException::forPath($this->getTargetPath());
```
webmozart: IMO this exception is misleading. The error is not that the target resource is not found, but that the link resource is not attached to a repository.

I think you should throw a RuntimeException here (and in similar cases).

webmozart: Also, the test for this is missing.

tgalopin:
I made the changes

tests/AbstractEditableRepositoryTest.php
```diff
@@ -473,4 +480,13 @@ public function testDirectoryLink()
         $this->assertSame($this->readRepo, $target->getRepository());
         $this->assertCount(2, $target->listChildren());
     }
+
+    /**
+     * @expectedException RuntimeException
```
webmozart: Can you adapt the type of the expected Exception?

webmozart:
Thanks @tgalopin! :)

PR #69, opened 2015-10-11
part of scrutinizer fixes

EmanueleMinotto:
Ref: https://github.com/puli/issues/issues/92

### DIFF
--- a/src/AbstractPathMappingRepository.php
+++ b/src/AbstractPathMappingRepository.php
@@ -208,7 +208,7 @@ abstract class AbstractPathMappingRepository extends AbstractRepository
     /**
      * Create a filesystem or generic resource.
      *
-     * @param string $filesystemPath
+     * @param string|null $filesystemPath
      *
      * @return DirectoryResource|FileResource|GenericResource
      */
@@ -234,8 +234,8 @@ abstract class AbstractPathMappingRepository extends AbstractRepository
     /**
      * Create a link resource to another resource of the repository.
      *
-     * @param string $targetPath The target path.
-     * @param string $path       The repository path.
+     * @param string      $targetPath The target path.
+     * @param string|null $path       The repository path.
      *
      * @return LinkResource The link resource.
      *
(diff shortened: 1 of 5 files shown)


src/AbstractPathMappingRepository.php
```diff
@@ -208,7 +208,7 @@ protected function sortStore()
     /**
      * Create a filesystem or generic resource.
      *
-     * @param string $filesystemPath
+     * @param string|bool|null $filesystemPath
```
webmozart: Why bool?

src/InMemoryRepository.php
```diff
@@ -81,7 +81,7 @@ public function find($query, $language = 'glob')
         $resources = array();
 
         if (Glob::isDynamic($query)) {
-            $resources = $this->getGlobIterator($query);
+            $resources = iterator_to_array($this->getGlobIterator($query));
```
webmozart: This isn't needed as the ArrayResourceCollection accepts iterators. You could add `Traversable` to the constructor argument's PHPDoc of that class instead.

src/InMemoryRepository.php
```diff
@@ -183,7 +183,7 @@ public function listChildren($path)
     {
         $iterator = $this->getChildIterator($this->get($path));
 
-        return new ArrayResourceCollection($iterator);
+        return new ArrayResourceCollection(iterator_to_array($iterator));
```
webmozart: same here

src/AbstractPathMappingRepository.php
EmanueleMinotto: Misunderstood a scrutinizer suggestion, fixed now

src/InMemoryRepository.php
EmanueleMinotto: ArrayResourceCollection currently requires both `Traversable` and `ArrayAccess` (afaik), but there isn't an interface extending both them (and the creation just to fill a PHPDoc seems useless to me), so I would suggest to revert this change and just ignore the Scrutinizer issue until next changes to the ArrayResourceCollection.

What to you think? Did I miss something?

webmozart: No, ArrayResourceCollection just requires any `Traversable` object - you can see that that object is passed to `replace()`, which in turn calls `iterator_to_array()`.

src/Resource/Collection/ArrayResourceCollection.php
```diff
@@ -40,7 +40,7 @@ class ArrayResourceCollection implements IteratorAggregate, ResourceCollection
      * You can pass the resources that you want to initially store in the
      * collection as argument.
      *
-     * @param PuliResource[] $resources The resources to store in the collection.
+     * @param PuliResource[]|\Traversable $resources The resources to store in the collection.
```
webmozart: Can you change this to `Traversable` (intead of `\Traversable`) and import the class instead? Also please add this as well for `replace()` and `merge()` (in the `ResourceCollection` interface).

PR #70, opened 2015-10-19
Mixed paths check

EmanueleMinotto:
Closes https://github.com/puli/issues/issues/133

### DIFF
--- a/src/PathMappingRepository.php
+++ b/src/PathMappingRepository.php
@@ -219,6 +219,10 @@ class PathMappingRepository extends AbstractPathMappingRepository implements Edi
                 return $this->resolveRelativePaths($filesystemPaths);
             }
 
+            if (null === $filesystemPaths && '/' !== $path) {
+                return $this->mergeMixedParent($path, $onlyFirst);
+            }
+
             return array(null);
         }
 
@@ -257,6 +261,42 @@ class PathMappingRepository extends AbstractPathMappingRepository implements Edi
         }
 
         return $this->resolveRelativePaths($filesystemPaths);
+    }
+
+    /**
+     * If the resource has a parent that's a reference to a directory, then
+     * I'll resolve it and append the filesystem part.
+     *
+     * @param string $path      The repository path.
+     * @param bool   $onlyFirst Should the method stop on the first path found?
+     *
+     * @return string[]|null[]
+     */
+    private function mergeMixedParent($path, $onlyFirst = true)
+    {
+        $resolvedFilesystemPaths = array();
+        $resolvedFilesystemPathsPart = array();
+        $parentPath = $path;
+
+        for ($i = 0; $i < substr_count($path, '/') - 1; ++$i) {
+            $parentPath = dirname($parentPath);
+
+            if ($this->store->exists($parentPath)) {
+                $resolvedFilesystemPathsPart = $this->resolveFilesystemPaths($parentPath, $onlyFirst);
+
+                foreach ($resolvedFilesystemPathsPart as $key => $value) {
+                    $resolvedFilesystemPathsPart[$key] = $value;
+                }
+            }
+        }
+
+        foreach ($resolvedFilesystemPathsPart as $key => $value) {
+            if (is_dir($value)) {
+                $resolvedFilesystemPaths[$key] = $value.substr($path, strlen($parentPath));
+            }
+        }
+
+        return $resolvedFilesystemPaths ?: array(null);
     }
 
     /**
--- a/tests/PathMappingRepositoryTest.php
+++ b/tests/PathMappingRepositoryTest.php
@@ -131,6 +131,103 @@ class PathMappingRepositoryTest extends AbstractPathMappingRepositoryTest
         $this->assertEquals(Path::normalize(__DIR__.'/Fixtures/dir5/sub'), $resource->getFilesystemPath());
     }
 
+    public function testMergeDirectParent()
+    {
+        $root = __DIR__.'/Fixtures/dir5';
+
+        $this->store->set('/', null);
+        $this->store->set('/webmozart', array($root));
+        $this->store->set('/webmozart/sub', null);
+        $this->store->set('/webmozart/sub/file1', array($root.'/file1'));
+
+        /*
+         * /webmozart/sub should have three children:
+         *      -   file1 from the virtual children
+         *          (directly resolve /webmozart/sub/file1 to /Fixtures/dir5/file1)
+         *      -   file3 and file4 from the real children
+         *          (resolve /webmozart to /Fixtures/dir5 and then list children of /Fixtures/dir5/sub)
+         */
+        $resources = $this->repo->listChildren('/webmozart/sub');
+        $this->assertCount(3, $resources);
+
+        $paths = $resources->getPaths();
+        sort($paths);
+
+        $this->assertSame('/webmozart/sub/file1', $paths[0]);
+        $this->assertSame('/webmozart/sub/file3', $paths[1]);
+        $this->assertSame('/webmozart/sub/file4', $paths[2]);
+    }
+
+    public function testMergeSubParents()
+    {
+        $root = __DIR__.'/Fixtures';
+
+        $this->store->set('/', null);
+        $this->store->set('/webmozart', array($root));
+        $this->store->set('/webmozart/dir5', null);
+        $this->store->set('/webmozart/dir5/sub', null);
+        $this->store->set('/webmozart/dir5/sub/file1', array($root.'/file1'));
+
+        /*
+         * /webmozart/dir5/sub should have three children:
+         *      -   file1 from the virtual children
+         *          (directly resolve /webmozart/dir5/sub/file1 to /Fixtures/dir5/file1)
+         *      -   file3 and file4 from the real children
+         *          (resolve /webmozart to /Fixtures and then list children of /Fixtures/dir5/sub)
+         */
+        $resources = $this->repo->listChildren('/webmozart/dir5/sub');
+        $this->assertCount(3, $resources);
+
+        $paths = $resources->getPaths();
+        sort($paths);
+
+        $this->assertSame('/webmozart/dir5/sub/file1', $paths[0]);
+        $this->assertSame('/webmozart/dir5/sub/file3', $paths[1]);
+        $this->assertSame('/webmozart/dir5/sub/file4', $paths[2]);
+    }
+
+    public function testMergeMultipleParents()
+    {
+        $root = __DIR__.'/Fixtures';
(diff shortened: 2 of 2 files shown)


EmanueleMinotto:
@tgalopin let me know if this could be a solution and if that test is covering the use case reported in the issue. Thanks :)

src/PathMappingRepository.php
```diff
@@ -260,6 +264,39 @@ private function resolveFilesystemPaths($path, $onlyFirst = true)
     }
 
     /**
+     * If the resource has a parent that's a symbolic link to a directory, then
```
tgalopin: Technically speaking, it's not a symbolic link :) .

src/PathMappingRepository.php
```diff
@@ -219,6 +219,10 @@ private function resolveFilesystemPaths($path, $onlyFirst = true)
                 return $this->resolveRelativePaths($filesystemPaths);
             }
 
+            if (null === $filesystemPaths && '/' !== $path) {
```
tgalopin: Why `'/' !== $path`?

tgalopin:
If I understand correctly, this PR will resolve only the first intermediate parent wouldn't it?

If the store contains:
``` text
/path => [/path1, /path2]
/path/sub1/sub2 => /path3
```

Then the path `/path/sub1/sub2` could be either:
``` text
/path3/file
/path1/sub1/sub2/file
/path2/sub1/sub2/file 
```

And your PR will only try `/path3/file` from what I se, wouldn't it?

I think a more formally defined algorithm in two steps would be clearer and more powerful :) :
1. Find possible paths (`/path3/file`, `/path1/sub1/sub2/file`, `/path2/sub1/sub2/file`)
2. Return the first available

Your PR is a great basis for this but I think we should reorganize it a bit :) .

src/PathMappingRepository.php
```diff
@@ -260,6 +264,39 @@ private function resolveFilesystemPaths($path, $onlyFirst = true)
...
+        $resolvedFilesystemPaths = array(null);
+        $parentPath = $path;
+
+        for ($i = 0; $i < substr_count($path, '/') - 1; ++$i) {
+            $parentPath = dirname($parentPath);
+
+            if ($this->store->exists($parentPath)) {
+                $resolvedFilesystemPaths = $this->resolveFilesystemPaths($parentPath, $onlyFirst);
+
+                break;
```
tgalopin: I don't think you should beark here: you should break when you resolved a valid resource, ie when `resolvedFilesystemPaths` returned something valid :) . Technically, you can have multiple parents.

tgalopin: Oh and by the way, you can also use `$onlyFirst` here to break only when the developer want to get the first valid resource. If he does not (if `$onlyFirst === false`) you can resolve all the valid parent paths.

tests/PathMappingRepositoryTest.php
```diff
@@ -131,6 +131,29 @@ public function testResolveFilesystemResource()
...
+        // Get
+        $resource = $this->repo->get('/sub1/sub2/sub3');
+
+        $this->assertInstanceOf('Puli\Repository\Api\Resource\FilesystemResource', $resource);
+        $this->assertEquals(Path::normalize(__DIR__.'/Fixtures/dir6/sub1/sub2/sub3'), $resource->getFilesystemPath());
+
+        // Find
+        $children = $this->repo->listChildren('/sub1/sub2/sub3');
+
+        $this->assertCount(2, $children);
```
tgalopin: This test is valid but I think you don't take into account the possibility of multiple parents.

Imagine this store:
``` text
/app => /dir1
/app/blog => /dir2
/app/blog/res => /dir3
```

When you access `/app/blog/res/views/index.html`, the test should check, in this order, all the following paths:

```
/dir3/views/index.html
/dir2/res/views/index.html
/dir1/blog/res/views/index.html
```

And return the first valid one if `$onlyFirst == true` or the list of all of them otherwise.

src/PathMappingRepository.php
EmanueleMinotto: Considering https://github.com/puli/repository/pull/70/files#diff-9eba964c7c0df37591f1971e96d907c4R280 the `/` path shouldn't be considered at all, so I'm adding it to the condition to prevent useless execution

EmanueleMinotto:
@tgalopin thanks for all your feedbacks, can you check now please?

src/PathMappingRepository.php
tgalopin: Okay, thanks for the explaination :) .

tests/PathMappingRepositoryTest.php
```diff
@@ -131,6 +131,38 @@ public function testResolveFilesystemResource()
         $this->assertEquals(Path::normalize(__DIR__.'/Fixtures/dir5/sub'), $resource->getFilesystemPath());
     }
 
+    public function testMergeMixedParent()
```
tgalopin: I'm not sure but I think you are still missing a test case here. If you're resolving `/sub1/sub2/sub3`, there are a few possibilities:

- `/sub1/sub2/sub3` exists (tested in other tests)
- `/sub1/sub2/sub3` doesn't exists but `/sub1/sub2` exists => you check the resource `/dir2/sub3`
- neither `/sub1/sub2/sub3` nor `/sub1/sub2` exist but `/sub1` exists => you check the resource `/dir1/sub2/sub3`

I think you didn't check all these possibilities. If you want, I can write the test I have in mind in a gist. Perhaps am I misunderstanding your code though, if you think I'm wrong please tell me :) !

EmanueleMinotto: @tgalopin yes please ^^' it would be great to have that test, thanks

tgalopin: I'll try that then :) .

EmanueleMinotto:
@tgalopin I’ve replaced my test with your tests. Am I missing anything else?

tests/PathMappingRepositoryTest.php
```diff
@@ -131,6 +131,79 @@ public function testResolveFilesystemResource()
...
+
+        /*
+         * /webmozart/sub should have three children:
+         *      -   file1 from the virtual children
+         *          (directly resolve /webmozart/sub/file1 to /Fixtures/dir5/file1)
+         *      -   file3 and file4 from the real children
+         *          (resolve /webmozart to /Fixtures/dir5 and then list children of /Fixtures/dir5/sub)
+         */
+        $resources = $this->repo->listChildren('/webmozart/sub');
+        $this->assertCount(3, $resources);
```
tgalopin: Can you check that the resources are always in the same order? Something like testing `$resources[0]`, `$resources[1]` and `$resources[2]` for their path. That would prevent problems in the future.

tests/PathMappingRepositoryTest.php
```diff
@@ -131,6 +131,79 @@ public function testResolveFilesystemResource()
...
+         * children of /webmozart/sub shoudl return:
+         *      /Fixtures/dir3/sub/file1
+         *      /Fixtures/dir4/sub/file2 (override /Fixtures/dir3/sub/file2)
+         *      /Fixtures/dir5/sub/file3 (override /Fixtures/dir4/sub/file3)
+         *      /Fixtures/dir5/sub/file4
+         *      /Fixtures/dir1/file1 (from first virtual file)
+         *      /Fixtures/dir1/file2 (from second virtual file)
+         */
+        $resources = $this->repo->listChildren('/webmozart/sub');
+        $this->assertCount(6, $resources);
```
tgalopin: Same here.

tests/PathMappingRepositoryTest.php
```diff
@@ -131,6 +131,79 @@ public function testResolveFilesystemResource()
...
+
+        /*
+         * /webmozart/dir5/sub should have three children:
+         *      -   file1 from the virtual children
+         *          (directly resolve /webmozart/dir5/sub/file1 to /Fixtures/dir5/file1)
+         *      -   file3 and file4 from the real children
+         *          (resolve /webmozart to /Fixtures and then list children of /Fixtures/dir5/sub)
+         */
+        $resources = $this->repo->listChildren('/webmozart/dir5/sub');
+        $this->assertCount(3, $resources);
```
tgalopin: Same here.

tgalopin:
Awesome! Once you add the resource order test, I think this will be ready to merge :) !

EmanueleMinotto:
Added the assertions about the order, but I'm not sure if the order of `testMergeDirectParent` and `testMergeSubParents` results are right.
@tgalopin are them right or should I fix them?

tgalopin:
That's a good question. @webmozart should we sort children by name? I think we shouldn't, it should be taken care of by the developer, not by the library, as the filsystem iterators in PHP. What do you think?

webmozart:
The order of children is usually undefined. You can sort them in the test case before you compare them.

EmanueleMinotto:
@webmozart is the results order required to complete this PR?

tgalopin:
I think you should just order the resources by path in the tests before checking the resources themselves. It would allow you to check resources are the right ones while not fixing an abritrary order.

EmanueleMinotto:
@tgalopin I've updated the tests, is that what you mean?

tgalopin:
Yes it is. I think it's ready for merge @webmozart. Thanks very much for your work @EmanueleMinotto, that is a great PR!

EmanueleMinotto:
great! @tgalopin thank you for your feedbacks :)

tests/PathMappingRepositoryTest.php
```diff
@@ -131,6 +131,103 @@ public function testResolveFilesystemResource()
         $this->assertEquals(Path::normalize(__DIR__.'/Fixtures/dir5/sub'), $resource->getFilesystemPath());
     }
 
+    public function testMergeDirectParent()
```
webmozart: Can this test be moved to the abstract repository test? The same functionality should be provided by any repository implementation.

tgalopin: This one use the store so it's too specific to this repo IMO. But I agree we should test this feature the same way everywhere, there is a bit of work here.

webmozart: You could test the same functionality by adding `LinkResource` instances, no?

tgalopin: I think we should isolate fetching from adding shouldn't we? If the `add()` solve a problem by advance, we could miss it in the fetching.

tgalopin:
I think we should merge that and implement the test in other repositories afterwards. It would fix https://github.com/puli/issues/issues/133 and we could move forward.

webmozart:
Alright then! Thanks :)

PR #72, opened 2015-11-10
Fetching subdirectories of LinkResource

msojda:
Fixes puli/issues#129. 

As @tgalopin suggested I'm creating PR for fetching linked resources only.

### DIFF
--- a/src/InMemoryRepository.php
+++ b/src/InMemoryRepository.php
@@ -289,7 +289,7 @@ class InMemoryRepository extends AbstractRepository implements EditableRepositor
     /**
      * Returns link from requested path if any found.
      *
-     * @param $path The path.
+     * @param string $path The path.
      *
      * @return LinkResource The link.
      */
--- a/src/InMemoryRepository.php
+++ b/src/InMemoryRepository.php
@@ -19,6 +19,7 @@ use Puli\Repository\Api\ResourceNotFoundException;
 use Puli\Repository\Api\UnsupportedResourceException;
 use Puli\Repository\Resource\Collection\ArrayResourceCollection;
 use Puli\Repository\Resource\GenericResource;
+use Puli\Repository\Resource\LinkResource;
 use Webmozart\Assert\Assert;
 use Webmozart\Glob\Glob;
 use Webmozart\Glob\Iterator\GlobFilterIterator;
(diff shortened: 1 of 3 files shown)


src/InMemoryRepository.php
```diff
@@ -63,11 +64,17 @@ public function get($path)
...
 
-        if (!isset($this->resources[$path])) {
-            throw ResourceNotFoundException::forPath($path);
+        if (isset($this->resources[$path])) {
+            return $this->resources[$path];
         }
 
-        return $this->resources[$path];
+        if ($link = $this->getLinkFromPath(Path::getDirectory($path))) {
+            $relativeTargetPath = ltrim($path, $link->getPath());
```
EmanueleMinotto: The second parameter is a set of characters, this won't give the relative path.

src/InMemoryRepository.php
```diff
@@ -281,6 +288,28 @@ private function getChildIterator(PuliResource $resource)
...
+            return;
+        }
+
+        if (!isset($this->resources[$path])) {
+            if ($link = $this->getLinkFromPath(Path::getDirectory($path))) {
+                return $link;
+            }
+        } elseif (isset($this->resources[$path]) && $this->resources[$path] instanceof LinkResource) {
+            return $this->resources[$path];
+        }
```
EmanueleMinotto: You can reduce the complexity replacing this block with:
```php
if (!isset($this->resources[$path]) && $link = $this->getLinkFromPath(Path::getDirectory($path))) {
    return $link;
}

if (isset($this->resources[$path]) && $this->resources[$path] instanceof LinkResource) {
    return $this->resources[$path];
}
```

EmanueleMinotto: You can reduce the complexity replacing with this block:
```php
if (isset($this->resources[$path]) && $this->resources[$path] instanceof LinkResource) {
    return $this->resources[$path];
}

if (!isset($this->resources[$path]) && $link = $this->getLinkFromPath(Path::getDirectory($path))) {
    return $link;
}
```

src/InMemoryRepository.php
```diff
@@ -281,6 +288,28 @@ private function getChildIterator(PuliResource $resource)
     }
 
     /**
+     * Returns link from requested path if any found.
+     *
+     * @param string $path The path.
+     *
+     * @return LinkResource The link.
```
EmanueleMinotto: If `'/' === $path` this method returns `null`, so I think should be added here.

src/InMemoryRepository.php
tgalopin: You can use `Path::makeRelative()`

webmozart:
Thanks @msojda for working on this! :) Please don't forget to add tests that cover the different cases and edge cases of this functionality. Ideally, you should add these tests to the abstract test classes, so that the same tests are used for all repository implementations.

webmozart:
What's the status here?

tgalopin:
This aimed to fix https://github.com/puli/issues/issues/129 but became a dupplicate of https://github.com/puli/repository/pull/64 when we chose to not implement add and removal in links children. I think this can be closed, but it depends on the amount of work you did there @msojda : did you work a lot on fetching resources? It could be interesting to see how you did it and integrate some ideas in https://github.com/puli/repository/pull/64.

PR #74, opened 2015-12-17
Add Symfony 3 filesystem component in composer.json

SenseException:
After the release of Symfony 3 and the fact, that puli/repository is using the filesystem component, I added Symfony 3 support in composer.json.


webmozart:
Thanks @SenseException! :)

PR #75, opened 2015-12-17
Added a JSON schema for the generated mapping files

webmozart:
I created a JSON schema for the mapping files generated by the path mapping repositories. In the future, these repositories should make sure to comply to the schema so that we can safely consume mapping files with other tools, such as a JavaScript library.

ping @tgalopin


tgalopin:
Great step forward, thanks :) ! Do you think I should change the internal storage to comply to this or I should create a method to dump the internal storage to this?

PR #79, opened 2015-12-30
Refactored JsonRepository

webmozart:


composer.json
```diff
@@ -12,18 +12,18 @@
     "require": {
         "php": "^5.3.9|^7.0",
         "webmozart/path-util": "^2.2",
-        "webmozart/glob": "^4.0",
+        "webmozart/glob": "^4.1",
         "webmozart/assert": "^1.0"
     },
     "require-dev": {
-        "webmozart/key-value-store": "^1.0.0-beta5",
```
tgalopin: I would not remove it as we will probably use stores in the ChangeStream. Or should we put it as a suggestion?

src/AbstractJsonRepository.php
```diff
@@ -0,0 +1,569 @@
...
+     *                                         Paths inside that directory are
+     *                                         stored as relative paths. Paths
+     *                                         outside that directory are stored
+     *                                         as absolute paths.
+     * @param ChangeStream|null $changeStream  If provided, the repository will
+     *                                         append resource changes to this
+     *                                         change stream.
+     * @param bool              $validateJson  Whether to validate the JSON file
+     *                                         against the schema. Slow but
+     *                                         spots problems.
```
tgalopin: What do you think of enabling it by default on dev repository and disabling it by default on prod one.

webmozart: I don't think enabling it by default makes much sense. The flag has two main purposes:

* spotting in the automatic tests whether the repository generates invalid JSON
* spotting at runtime whether the input JSON is valid *if* the JSON is modified externally (by the user or another tool)

Since in most cases the repository will be used to generate the JSON, we can assume that the JSON is valid IMO.

composer.json
webmozart: You can add it back in in the respective PR, both in "require-dev" and "suggest" (like in puli/discovery).

webmozart:
I removed the ChangeStream from JsonRepository now. The problem is that we can't really detect changes with the ChangeStream, since we don't traverse the filesystem (like in OptimizedJsonRepository) and add the child nodes to the stream.

Instead, I used the logic that was already in JsonRepository to support building of ResourceStack objects. An additional JSON entry "_order" was necessary to fix cases where child paths where overridden by parent paths. See the test case `AbstractEditableRepositoryTest::testOverrideSuperPath()` for an example.

tgalopin:
I don't understand the problem with the change stream. Why can't we really detect changes with it?

webmozart:
Because we're only adding the mappings themselves to the changestream, but not the files in the mapped directories. Imagine you do:

~~~php
$repo->add('/path', new DirectoryResource('/dir1'));
$repo->add('/path', new DirectoryResource('/dir2'));
~~~

Now you can build the stack for `/path` with the ChangeStream, but not for e.g. `/path/file`, since from the perspective of the ChangeStream that path does not exist.

src/AbstractJsonRepository.php
```diff
@@ -0,0 +1,595 @@
...
+use Puli\Repository\Resource\GenericResource;
+use Puli\Repository\Resource\LinkResource;
+use RuntimeException;
+use Webmozart\Assert\Assert;
+use Webmozart\Json\JsonDecoder;
+use Webmozart\Json\JsonEncoder;
+use Webmozart\PathUtil\Path;
+
+/**
+ * Abstract base for Path mapping repositories.
```
tgalopin: You should change this comment :) .

src/AbstractJsonRepository.php
```diff
@@ -0,0 +1,595 @@
...
+    }
+
+    /**
+     * Returns the short name of a fully-qualified class name.
+     *
+     * @param string $className The fully-qualified class name.
+     *
+     * @return string The short class name.
+     */
+    private function getShortClassName($className)
```
tgalopin: Is the short class name only used for exceptions messages?

src/JsonRepository.php
```diff
@@ -0,0 +1,932 @@
...
+use Puli\Repository\Api\Resource\PuliResource;
+use Puli\Repository\Api\ResourceNotFoundException;
+use Puli\Repository\ChangeStream\ResourceStack;
+use RecursiveIteratorIterator;
+use Webmozart\Glob\Glob;
+use Webmozart\Glob\Iterator\RecursiveDirectoryIterator;
+use Webmozart\PathUtil\Path;
+
+/**
+ * A development path mapping resource repository.
```
tgalopin: I think you should change this comment as well.

src/JsonRepository.php
```diff
@@ -0,0 +1,932 @@
...
+ * Resources can be added with the method {@link add()}:
+ *
+ * ```php
+ * use Puli\Repository\JsonRepository;
+ *
+ * $repo = new JsonRepository();
+ * $repo->add('/css', new DirectoryResource('/path/to/project/res/css'));
+ * ```
+ *
+ * This repository only supports instances of FilesystemResource.
```
tgalopin: This is not true anymore btw, we support links.

src/JsonRepository.php
```diff
@@ -0,0 +1,932 @@
...
+        krsort($this->json);
+
+        // If a mapping exists for a sub-path of this resource
+        // (e.g. $path = /a, mapped sub-path = /a/b)
+        // we need to record the order, since by default sub-paths are
+        // preferred over super paths
+
+        $references = $this->searchReferences(
+            $path,
+            // Don't stop for the first result
```
tgalopin: What about using constants instead, for better readability with less comments?

src/JsonRepository.php
```diff
@@ -0,0 +1,932 @@
...
+
+        foreach ($checkResults as $path => $filesystemPath) {
+            if (!array_key_exists($path, $this->json)) {
+                $nonDeletablePaths[] = $filesystemPath;
+            }
+        }
+
+        if (count($nonDeletablePaths) === 1) {
+            throw new BadMethodCallException(sprintf(
+                'The remove query "%s" matched a resource that is not a path mapping', $glob
```
tgalopin: As we know we have only one, we could display it's path for DX.

webmozart: Good idea!

src/AbstractJsonRepository.php
webmozart: Yes

src/JsonRepository.php
webmozart: Done in 7e73126

src/JsonRepository.php
webmozart: Fixed

src/JsonRepository.php
webmozart: Done

src/AbstractJsonRepository.php
webmozart: Thanks, done :)

src/OptimizedJsonRepository.php
```diff
@@ -0,0 +1,222 @@
...
+ * A repository backed by a JSON file optimized for reading.
+ *
+ * The generated JSON file is described by res/schema/repository-schema-1.0.json.
+ *
+ * Resources can be added with the method {@link add()}:
+ *
+ * ```php
+ * use Puli\Repository\JsonRepository;
+ *
+ * $repo = new JsonRepository('/path/to/repository.json', '/path/to/project');
```
tgalopin: Technically, here, it's `OptimizedJsonRepository` :) .

src/OptimizedJsonRepository.php
```diff
@@ -0,0 +1,222 @@
...
+        // We're only interested in the first entry of eventual arrays
+        if (is_array($reference)) {
+            $reference = reset($reference);
+        }
+
+        if ($this->isFilesystemReference($reference)) {
+            $reference = Path::makeAbsolute($reference, $this->baseDirectory);
+
+            // Ignore non-existing files. Not sure this is the right
+            // thing to do.
```
tgalopin: I think we should throw an expection here. I think that returning an empty array would most of the time results in a `ResourceNotFound` exception, which is misleading as it's not the real problem.

tests/AbstractEditableRepositoryTest.php
```diff
@@ -288,10 +291,352 @@ public function testAddExpectsResource()
...
+        $v2 = $stack->get(2);
+
+        $this->assertInstanceOf('Puli\Repository\Api\Resource\BodyResource', $v2);
+        $this->assertSame('/webmozart/file', $v2->getPath());
+        $this->assertSame($this->readRepo, $v2->getRepository());
+        $this->assertSame('BODY2', $v2->getBody());
+    }
+
+    public function testOverrideSuperPath()
+    {
```
tgalopin: That's a great test case I didn't think of. Does it pass?

tests/AbstractJsonRepositoryTest.php
```diff
@@ -0,0 +1,154 @@
...
+     * @var int
+     */
+    protected static $nextFileId = 0;
+
+    /**
+     * Counter to avoid collisions during tests on directories.
+     *
+     * @var int
+     */
+    protected static $nextDirectoryId = 0;
```
tgalopin: What do you think of puttinh this as private ?

tests/JsonRepositoryTest.php
```diff
@@ -0,0 +1,59 @@
...
+ * @author Titouan Galopin <galopintitouan@gmail.com>
+ */
+class JsonRepositoryTest extends AbstractJsonRepositoryTest
+{
+    /**
+     * Counter to avoid collisions during tests on directories.
+     *
+     * @var int
+     */
+    protected static $nextDirectoryId = 0;
```
tgalopin: Why do you need this here?

webmozart: Good question, you introduced this I think :)

tests/AbstractEditableRepositoryTest.php
webmozart: Yes of course :)

src/OptimizedJsonRepository.php
webmozart: Thanks :) Fixed

tests/JsonRepositoryTest.php
webmozart: Removed

tests/AbstractJsonRepositoryTest.php
webmozart: Good idea, done

src/OptimizedJsonRepository.php
webmozart: I logged a warning now in this case. I don't want to fail completely because then find() will fail even if just one reference is configured with a wrong path. Since references could be configured in other packages that the user has no control over, failing completely is not very user friendly.

webmozart:
I pushed another update. Warnings are passed to a logger now if references are not found. I improved the ChangeStream naming, implementation and tests along the go.

composer.json
```diff
@@ -12,18 +12,19 @@
...
         "webmozart/path-util": "^2.2",
-        "webmozart/glob": "^4.0",
-        "webmozart/assert": "^1.0"
+        "webmozart/glob": "^4.1",
+        "webmozart/assert": "^1.0",
+        "psr/log": "^1.0"
     },
     "require-dev": {
-        "webmozart/key-value-store": "^1.0.0-beta5",
+        "webmozart/json": "^1.2@dev",
```
tgalopin: What do you think of releasing a beta of webmozart/json and use it here? I think it would be great for beta10 to be as stable as possible.

src/ChangeStream/InMemoryChangeStream.php
```diff
@@ -11,53 +11,83 @@
...
 
     /**
      * {@inheritdoc}
      */
-    public function buildStack(ResourceRepository $repository, $path)
+    public function purge($path)
     {
-        if (!isset($this->stack[$path])) {
-            throw new InvalidArgumentException(sprintf('No version of path %s were found in the ChangeStream', $path));
+        unset($this->versions[$path]);
```
tgalopin: Wouldn't this throw a notice if the path is invalid? I would test it before.

composer.json
webmozart: Actually, 1.2 is already released. I'll change this.

src/ChangeStream/InMemoryChangeStream.php
webmozart: No, `unset()` never fails.

PR #84, opened 2016-01-06
Run --prefer-lowest build on PHP 7

webmozart:


.travis.yml
```diff
@@ -12,12 +12,12 @@ matrix:
...
     - php: 5.5
     - php: 5.6
-    - php: 5.6
-      env: COMPOSER_FLAGS='--prefer-lowest --prefer-stable'
     - php: hhvm
     - php: nightly
     - php: 7.0
       env: COVERAGE=yes
+    - php: 7.0
+      env: COMPOSER_FLAGS='--prefer-lowest --prefer-stable'
```
xabbuh: What about merging this job with the coverage one?

webmozart: These are two different jobs. One with the current dependencies (on PHP 7), and another with the lowest one.

nicolas-grekas:
On Symfony, the benefit of doing `--prefer-lowest` on the highest PHP version is that it verifies that lowest versions of deps are somewhat  compatible with the latest PHP.
:+1:

webmozart:
I think it's useful to know whether a failing test results from PHP 7 incompatibility *or* from a wrong version constraint in composer.json, so I think we should keep the two separated for now. The build time of each is quite short, so it shouldn't be a problem.

PR #86, opened 2016-01-06
Moved the GitHub OAuth token to environment variable

webmozart:


EmanueleMinotto:
Without the token but forcing to `--prefer-dist` there will be always delays in forks, what about adding this?

```yaml
before_install:
  - composer config preferred-install source
  - if [ -n $COMPOSER_AUTH_TOKEN ]; then composer config github-oauth.github.com ${COMPOSER_AUTH_TOKEN}; fi
  - if [ -n $COMPOSER_AUTH_TOKEN ]; then composer config preferred-install dist; fi

install:
  - composer install --no-interaction --no-progress
```

webmozart:
Good idea!

webmozart:
Hm there's a problem. Apparently environment variables are not made available for PRs (due to security reasons), but that's the whole point of having a CI - testing whether PRs work before merging them.

EmanueleMinotto:
Yep, seems that the only way to speed up PRs from forks on CI systems is using a shared token.

Anyway without xdebug in PHP 5.3, tests take only 50 secs, this seems a reasonable time to me even if there isn't a global cache for Composer.

For the "need for speed", the cache could be used transparently using the lines above (and suggesting to apply the same thing in forks).

Just my opinion, of course. :)

PR #91, opened 2016-01-19
Fix real files removal in FilesystemRepository::clear with symlinks enabled

tgalopin:
Fix puli/issues#178

The `build` cli command do two things: clear the repository and rebuild it. To clean the repository, the manager call the method `clear` of the configured repository (in our case FilsystemRepository).

In the FilesystemRepository, `clear` calls `removeResource` on all the resources in the repository. When using the FilesystemRepository with symlinks, it means `removeResource` is called on all the links present in `.puli/path-mappings`. However, `removeResource` is, for the moment, always recursive: all the children of the given path will be removed before trying to remove the path itself.

When trying to remove a directory link, `removeResource` list all its children to remove them: however, the children of a directory link are the real children in the filesystem. So the real files are removed.

This PR fixes this.

### DIFF
--- a/src/FilesystemRepository.php
+++ b/src/FilesystemRepository.php
@@ -378,7 +378,7 @@ class FilesystemRepository extends AbstractEditableRepository
 
         ++$removed;
 
-        if (is_dir($filesystemPath)) {
+        if (is_dir($filesystemPath) && !is_link($filesystemPath)) {
             $iterator = $this->getDirectoryIterator($filesystemPath);
 
             foreach ($iterator as $childFilesystemPath) {
--- a/tests/AbstractFilesystemRepositorySymlinkTest.php
+++ b/tests/AbstractFilesystemRepositorySymlinkTest.php
@@ -1,0 +1,78 @@
+<?php
+
+/*
+ * This file is part of the puli/repository package.
+ *
+ * (c) Bernhard Schussek <bschussek@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Puli\Repository\Tests;
+
+use Puli\Repository\Resource\DirectoryResource;
+use Symfony\Component\Filesystem\Filesystem;
+use Webmozart\Glob\Test\TestUtil;
+
+/**
+ * @author Titouan Galopin <galopintitouan@gmail.com>
+ */
+abstract class AbstractFilesystemRepositorySymlinkTest extends AbstractFilesystemRepositoryTest
+{
+    protected $tempBaseDir;
+
+    protected $tempDir;
+
+    /**
+     * Copy fixtures to temporary directory to prevent messing up the real
+     * fixtures when symlinks do not work.
+     */
+    protected $tempFixtures;
+
+    protected function setUp()
+    {
+        $this->markAsSkippedIfSymlinkIsMissing();
+
+        $this->tempBaseDir = TestUtil::makeTempDir('puli-repository', __CLASS__);
+
+        // Create both directories in the same directory, so that relative links
+        // work from one to the other
+        $this->tempDir = $this->tempBaseDir.'/workspace';
+        $this->tempFixtures = $this->tempBaseDir.'/fixtures';
+
+        mkdir($this->tempDir);
+        mkdir($this->tempFixtures);
+
+        $filesystem = new Filesystem();
+        $filesystem->mirror(__DIR__.'/Fixtures', $this->tempFixtures);
+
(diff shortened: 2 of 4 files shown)


tests/AbstractFilesystemSymlinkRepositoryTest.php
```diff
@@ -0,0 +1,78 @@
...
+namespace Puli\Repository\Tests;
+
+use Puli\Repository\Resource\DirectoryResource;
+use Symfony\Component\Filesystem\Filesystem;
+use Webmozart\Glob\Test\TestUtil;
+
+/**
+ * @author Titouan Galopin <galopintitouan@gmail.com>
+ */
+abstract class AbstractFilesystemSymlinkRepositoryTest extends AbstractFilesystemRepositoryTest
```
webmozart: I would name this `AbstractFilesystemRepositorySymlinkTest` to be consistent with the other names:

* We are creating an `Abstract` class
* for the `FilesystemRepository`
* with `Symlink` enabled

webmozart:
Thanks :) The Windows tests fail, can you check?

tgalopin:
I saw that. I think the issue already existed, but I'm not sure. I'll try to fix this, but I'll do it in another PR :) .

webmozart:
I would prefer if this could be fixed here. [The current AppVeyor build is green](https://ci.appveyor.com/project/webmozart/repository/branch/1.0).

tgalopin:
Okay, I'll work here then :) .

tests/AbstractFilesystemSymlinkRepositoryTest.php
tgalopin: Okay :)

pkruithof:
Is there a workaround for this issue until this PR is merged? Our frontend is not working because of this :(

tgalopin:
You can add manually ` && !is_link($filesystemPath)` in the FilesystemRepository as a temporary solution. I'm having trouble to fix the tests under Windows.

webmozart:
@pkruithof Did you try switching to the JSON repository?

```
$ puli config repository.type json
```

or 

```
$ puli config --reset repository.type
```

tests/AbstractFilesystemSymlinkRepositoryTest.php
webmozart: Can you do this change?

pkruithof:
@webmozart that works, thanks!

tgalopin:
I *finally* found and fixed the problem ... Windows is really a mess. For your information: when you use DirectoryIterator on a directory link, under Unix you get resolved paths (under the path of the target of the link) whereas in Windows, the links are relative to the current link (even though these files does not exist). Anyway, this should be soon ready.

webmozart:
Thanks @tgalopin! :)

pkruithof:
@webmozart the JSON workaround no longer works for us, and we haven't updated Puli since my last comment :(

Could you tag a new release? This is really inconvenient for us right now.

webmozart:
@pkruithof Can you explain why the JSON workaround doesn't work for you?

pkruithof:
It happens during `puli publish --install`. I think it's because of the server config:

```json
    "extra": {
        "servers": {
            "local": {
                "installer": "symlink",
                "document-root": "web"
            }
        }
    }
```

I've tried switching this to `copy`, but that doesn't fix it.

I'll try to find the cause of this. It's strange since I didn't change anything since I added the workaround.

webmozart:
Ah yes that makes sense. `puli publish --install` uses the FilesystemRepository internally. Thanks for the clarification.

pkruithof:
Is there another workaround for that or are you working on a new release?

pkruithof:
@webmozart sorry to bother you like this but I'm really stuck right now. I've tried all kinds of version constraints to roll back to a previous Puli version, but since we also use Symfony 3 this seems impossible.

A new beta release would really help us out, and since this is a pretty serious bug, I hope you agree it has some priority.

webmozart:
@pkruithof I'll look into releasing a fix release today. Sorry for your hassles! :(

webmozart:
@pkruithof Could you please verify that https://github.com/tgalopin/repository/commit/02a2dc0bda09fdda443660769a36381d9bd910ee fixes the problems for you? If you can, I'll release a new version with that commit applied.

pkruithof:
Confirmed, that fixes it.

webmozart:
Fix release here: https://github.com/puli/repository/releases/tag/1.0.0-beta10

pkruithof:
Sweet, thanks so much!

PR #95, opened 2016-09-29
Handle "non-real" paths

dantleech:
Use the resolved absolute path to determine the resource name.


webmozart:
Thanks! Could you fix the StyleCI issue?

